PR #33, opened 2021-11-19
ci: add build and publishing of operator artefacts

skabashnyuk:
### What does this PR do?
add build and publishing of operator artefacts
Reference implementation https://github.com/redhat-appstudio/application-service/blob/main/.github/workflows/build.yml
n/a

Publishing operator artifacts.

### How to test this PR?
n/a

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -12,7 +12,8 @@ endif
 # To re-generate a bundle for another specific version without changing the standard setup, you can:
 # - use the VERSION as arg of the bundle target (e.g make bundle VERSION=0.0.2)
 # - use environment variables to overwrite this value (e.g export VERSION=0.0.2)
-VERSION ?= 0.0.1
+VERSION ?= 0.1.0
+TAG_NAME ?= latest
 
 # CHANNELS define the bundle channels used in the bundle.
 # Add a new line here if you would like to change its default config. (E.g CHANNELS = "candidate,fast,stable")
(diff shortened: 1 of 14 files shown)


Makefile
```diff
@@ -12,7 +12,8 @@ endif
 # To re-generate a bundle for another specific version without changing the standard setup, you can:
 # - use the VERSION as arg of the bundle target (e.g make bundle VERSION=0.0.2)
 # - use environment variables to overwrite this value (e.g export VERSION=0.0.2)
-VERSION ?= 0.0.1
+VERSION ?= 0.1.0
+TAG_NAME ?= latest
```
sparkoo, line 16: how we will use `lates` tag? Is it latest released? Or next development version? If latter, shouldn't we use `next`?

Makefile
```diff
@@ -38,14 +39,14 @@ BUNDLE_METADATA_OPTS ?= $(BUNDLE_CHANNELS) $(BUNDLE_DEFAULT_CHANNEL)
 #
 # For example, running 'make bundle-build bundle-push catalog-build catalog-push' will build and push both
 # appstudio.redhat.org/service-provider-integration-operator-bundle:$VERSION and appstudio.redhat.org/service-provider-integration-operator-catalog:$VERSION.
-SPIO_IMAGE_TAG_BASE ?= appstudio.redhat.org/service-provider-integration-operator
+SPIO_IMAGE_TAG_BASE ?= quay.io/skabashn/service-provider-integration-operator
```
sparkoo, line 42: is this correct image?

config/manager/kustomization.yaml
```diff
@@ -5,6 +5,12 @@ generatorOptions:
...
-- name: manager-config
-  files:
+- files:
   - controller_manager_config.yaml
+  name: manager-config
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+images:
+- name: controller
+  newName: quay.io/skabashn/service-provider-integration-operator
```
sparkoo, line 15: is this correct image ?

Makefile
skabashnyuk: for now yes, eventually we will go to app-studio quay account.

config/manager/kustomization.yaml
skabashnyuk: for now yes, eventually we will go to app-studio quay account.

Makefile
skabashnyuk: tbh I'm not yet sure how it will work. it's par of originally generated code.
why do you proposed `next` I saw it here https://github.com/redhat-appstudio/application-service/blob/main/Makefile#L7 
Should I change it to `next`?

sparkoo: I don't know. As I understand it, `latest` is usually used for latest released version so if you use it, you will be on latest stable. `next` is used for cutting edge dev branch, that would mean latest commit in `master` branch. So it depends what we want to communicate with this name.

sparkoo (approved):

PR #35, opened 2021-11-22
feat: revised architecture based on SPIAccessTokenBinding and SPIAccessToken

metlos:
### What does this PR do?
This introduces the revised architecture using the `SPIAccessToken` and `SPIAccessTokenBinding`.

This illustrates how we could use webhooks to store the sensitive data into Vault and not in K8s. Webhooks are experimental in this picture and we might remove (parts of) them in the future.

Notably, this is still missing:
* reconciliation logic of `SPIAccessTokenBinding` not complete
* reconciliation of `SPIAccessToken` not complete
* access to Vault
* search logic for tokens


webhooks/old.go.bak
```diff
@@ -0,0 +1,65 @@
+// package webhooks
```
skabashnyuk, line 1: delete?

skabashnyuk (commented):
in general looks good

webhooks/old.go.bak
metlos: whoops

skabashnyuk (approved):

PR #53, opened 2022-01-11
feat: Svpi48 config

sparkoo:
### What does this PR do?
 - Extracts `baseUrl` and `sharedSecret` from config file into env
 - removes `kubeConfigPath` and use `-kubeconfig` parameter from operator-sdk
 - cleanup some unused code
 - updates deployment 
   - set `sharedSecret` under it's own key and mount it into containers as env variable
   - set `baseUrl` as env variable directly
   - remove `meta-config` and rather create secret directly
   - remove kustomize `vars`, that are soon to be deprecated and rather use `replacements`

https://issues.redhat.com/browse/SVPI-48

### How to test this PR?

To test, you have to use updated spi-oauth image that can run in this deployment `quay.io/mvala/spi-oauth:svpi48`. PR for it is here https://github.com/redhat-appstudio/service-provider-integration-oauth/pull/22. To complete that, I need to merge this first and update spi-operator dependency there.

spi-operator image here: `quay.io/mvala/spi-operator:svpi48`

Steps to test:
 1. update `config/k8s/ingress.yaml` with your kubernetes cluster hostname
 2. update `config/k8s/config.yaml` with your github oauth app values and your ingress host
 3. deploy with `make deploy_k8s SPIO_IMG=quay.io/mvala/spi-operator:svpi48-config SPIS_IMG=quay.io/mvala/spi-oauth:svpi48-config`


pkg/spi-shared/config/config_test.go
```diff
@@ -18,13 +18,15 @@ import (
...
 	"os"
 	"path/filepath"
-	"strings"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
 )
 
 func TestRead(t *testing.T) {
+	os.Setenv(sharedSecretEnv, "secretValue")
```
metlos, line 27: I'm not sure I like this.. I'm not a big fan of global variables... What about introducing these values back into the configuration and provide "setters" or at least documentation that these need to be set manually? That way this can be tested more cleanly and thread-safely without relying on values of global variables (which env is).

config/manager/manager.yaml
```diff
@@ -26,6 +26,13 @@ spec:
         env:
         - name: SPI_URL
```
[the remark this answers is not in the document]
sparkoo: I would rather do it in separate PR, it's in a lot of places and this PR is already bigger than I wanted.

skabashnyuk:
From the latest security best practices I've seen, it is common that sensitive data should not be exposed/configured as an environment variable. Could we change the way how `sharedSecret` is configured from an environment variable to a file? WDYT?

config/shared-config/kustomization.yaml
```diff
@@ -4,31 +4,9 @@
...
-    apiVersion: v1
-    name: oauth-metaconfig
-  fieldref:
-    fieldpath: data.oauth-host
+secretGenerator:
+  - name: oauth-config
+    files:
+    - config.yaml
+    literals:
+    - shared-secret=yaddayadda123$@#**
```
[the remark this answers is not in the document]
sparkoo: I don't know either. I've moved it into the file and load is same as `config.yaml`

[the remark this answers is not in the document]
sparkoo: I've moved it into `config.yaml` as it is now secret so we can have it in one.

Makefile
```diff
@@ -210,7 +210,7 @@ controller-gen: ## Download controller-gen locally if necessary.
 
 KUSTOMIZE = $(shell pwd)/bin/kustomize
 kustomize: ## Download kustomize locally if necessary.
-	$(call go-get-tool,$(KUSTOMIZE),sigs.k8s.io/kustomize/kustomize/v3@v3.8.7)
+	$(call go-get-tool,$(KUSTOMIZE),sigs.k8s.io/kustomize/kustomize/v4@v4.4.1)
```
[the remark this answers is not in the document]
sparkoo: Screenshot 2022-01-19 at 12 01 14

pkg/spi-shared/config/config_test.go
sparkoo: moved back to configfile

skabashnyuk (approved):
Works for me 
Знімок екрана 2022-01-19 о 14 07 18
Знімок екрана 2022-01-19 о 14 05 05
Знімок екрана 2022-01-19 о 14 04 57

config/k8s/config.yaml
```diff
@@ -1,9 +1,9 @@
-sharedSecretFile: $(CONFIG_MOUNT_PATH)/shared-secret
-baseUrl: $(OAUTH_HOST)
+sharedSecret: yaddayadda123$@#**
```
metlos, line 1: Is this dummy value required? Wouldn't something like `<REPLACE_ME_WITH_A_REAL_VALUE>` be more descriptive?

config/k8s/config.yaml
```diff
@@ -1,9 +1,9 @@
...
 serviceProviders:
 - type: GitHub
   clientId: "123"
   clientSecret: "42"
 - type: Quay
   clientId: "456"
   clientSecret: "54"
+baseUrl: http://oauth.192.168.64.46.nip.io
```
metlos, line 9: Is this dummy value required? Wouldn't something like `<REPLACE_ME_WITH_A_REAL_VALUE>` be more descriptive?

Also, we should be using `https` scheme here, IMHO.

config/k8s/ingress.yaml
```diff
@@ -4,7 +4,7 @@ metadata:
   name: oauth-ingress
 spec:
   rules:
-  - host: $(OAUTH_HOST)
+  - host: oauth.192.168.64.46.nip.io
```
metlos, line 7: Is this dummy value required? Wouldn't something like `<REPLACE_ME_WITH_A_REAL_VALUE>` be more descriptive?

main.go
```diff
@@ -98,15 +98,9 @@ func main() {
 		os.Exit(1)
 	}
 
-	pcfg, err := sharedConfig.LoadFrom(configFile)
+	cfg, err := sharedConfig.Config(configFile)
```
metlos, line 101: Why this rename? I won't fight to death over it, but `sharedConfig.LoadFrom` seems to describe what it is doing better than `sharedConfig.Config` IMHO. Since Go doesn't support method overloads, the best name IMHO would be `LoadFromFile` (if we ever want to add other ways of loading the configuration).

main.go
```diff
@@ -98,15 +98,9 @@ func main() {
...
 	if err != nil {
-		setupLog.Error(err, "failed to load the configuration file")
-		os.Exit(1)
-	}
-
-	cfg, err := pcfg.Inflate()
-	if err != nil {
-		setupLog.Error(err, "failed to initialize the configuration file")
+		setupLog.Error(err, "Failed to load the configuration")
```
metlos, line 103: Yes, the previous error message was stupid, but see? You're using "load" in the (much better) error message for a method called `Config` which supports my previous point about the rename from `LoadFrom` to `Config` 😉

metlos (commented):
In general this is looking good. Just a few minor stylistic comments...

config/k8s/config.yaml
sparkoo: yes, replaced with something that can be used lated as placeholder

config/k8s/config.yaml
sparkoo: replaced with placeholder

config/k8s/ingress.yaml
sparkoo: replaced with placeholder

main.go
sparkoo: I agree. I've renamed this in some phase of writing the code that `LoadFrom` did not make sense anymore. But now it does. Renamed back.

sparkoo:
@metlos I don't know about https. If this is only for local testing, I think it's not necessary. It will add more complexity with certificates and I'm not sure if oauth provider can redirect to https with self signed cert. Or can we use no cert with accepting the risk in browser? Anyway, I would solve it in separate PR.

metlos (approved):

PR #68, opened 2022-01-28
fix: Improve status handling in SPIAccessTokenBinding

metlos:
### What does this PR do?
Update the phase of the binding and also mirror the OAuthUrl of the token
directly in the status of the binding so that the users don't have to
read the token as well as the binding.

https://issues.redhat.com/browse/SVPI-44

### How to test this PR?
Follow the steps from the operator README to create SPIAccessTokenBinding etc. You should see the `phase` of the `SPIAccessTokenBinding` to have a non-empty value that reflects the phase of the token itself.

When the token is in the `AwaitingTokenData` phase, the binding should also be `AwaitingTokendData` phase. When the token is n the `Ready` phase, the binding should eventually transition to `Injected` phase once the secret is successfully created.

You should also see the OAuth URL in the status of the binding object, if the OAuth URL is required (i.e. when the phase is `AwaitingTokenData`).

### DIFF
--- a/api/v1beta1/spiaccesstoken_types.go
+++ b/api/v1beta1/spiaccesstoken_types.go
@@ -22,10 +22,12 @@ import (
 
 // SPIAccessTokenSpec defines the desired state of SPIAccessToken
 type SPIAccessTokenSpec struct {
+	//+kubebuilder:validation:Required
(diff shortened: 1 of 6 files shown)


skabashnyuk (approved):
Works for me
Знімок екрана 2022-01-31 о 13 26 03

PR #69, opened 2022-01-28
fix: svpi53 - spiaccesstoken lookup only namespace

sparkoo:
### What does this PR do?
When we're doing `SPIAccessToken` lookup, we're not limiting listing only single namespace where `SPIAccessTokenBinding` is. So we first find any token that is already on cluster and we think it's already there so we don't create new one. This patch limits lookup to only single namespace of `SPIAccessTokenBinding`.

https://issues.redhat.com/browse/SVPI-53

### How to test this PR?

image: `quay.io/mvala/spi-operator:53`

1. create 2 `SPIAccessTokenBinding` in different namespaces
2. without patch, the 2nd one did not create `SPIAccessToken`. With patch, it creates it.

### DIFF
--- a/pkg/serviceprovider/github.go
+++ b/pkg/serviceprovider/github.go
@@ -79,7 +79,10 @@ func (g *Github) LookupToken(ctx context.Context, cl client.Client, binding *api
 	// for now just return the first SPIAccessToken that we find so that we prevent infinitely many SPIAccessTokens
 	// being created during the tests :)
 	ats := &api.SPIAccessTokenList{}
-	if err := cl.List(ctx, ats, client.Limit(1)); err != nil {
+	if err := cl.List(ctx, ats, &client.ListOptions{
+		Namespace: binding.Namespace,
+		Limit:     1,
+	}); err != nil {
 		return nil, err
 	}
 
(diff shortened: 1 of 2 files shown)


pkg/serviceprovider/github.go
```diff
@@ -79,7 +79,10 @@ func (g *Github) LookupToken(ctx context.Context, cl client.Client, binding *api
 	// for now just return the first SPIAccessToken that we find so that we prevent infinitely many SPIAccessTokens
 	// being created during the tests :)
 	ats := &api.SPIAccessTokenList{}
-	if err := cl.List(ctx, ats, client.Limit(1)); err != nil {
+	if err := cl.List(ctx, ats, &client.ListOptions{
```
metlos, line 82: I think the canonical way of doing this is:
```
cl.List(ctx, ats, client.Limit(1), client.InNamespace(binding.Namespace))
```

metlos (approved):
Good catch!

pkg/serviceprovider/github.go
metlos: Or not, doesn't really matter...

skabashnyuk (approved):

PR #74, opened 2022-02-15
feat: SVPI-61 - vault deploy, init and configure

sparkoo:
### What does this PR do?
vault deployment and configuration on minikube or openshift

 - minikube `make deploy_minikube`
 - openshift `make deploy`
 - script to init vault 
   - initialize vault and stores unseal keys into `spi-vault-keys` secret
   - setup kubernetes auth
   - setup audit log
   - setup `spi` kv secret engine
 - automatically unseal vault at pod startup
 - script for root access
 - generate new deployment yamls with `./hack/vault-generate-template.sh`
 - updates README

I've tested on minikube, openshift deployed with `make deploy`, openshift with infra-deployments

https://issues.redhat.com/browse/SVPI-61

### How to test this PR?

1. have `kubectl` configured for desired cluster, openshift or minikube
2. run `make deploy_minikube` or `make deploy` for openshift
3. Vault should be initialized and configured. You can go to Vault shell, run `vault status`. Should be initialized and unsealsed
4. To gain root access on vault pod, run `sh /vault/userconfig/scripts/root.sh`. This generates root token and login with it.


hack/vault-init.sh
```diff
@@ -0,0 +1,157 @@
...
+
+init
+secret
+unseal
+ensureRootToken
+login
+audit
+k8sAuth
+spiSecretEngine
+restart
```
skabashnyuk, line 157: do we need to keep logged in state after restart?

config/vault/openshift/deployment.yaml
```diff
@@ -0,0 +1,359 @@
...
+              valueFrom:
+                fieldRef:
+                  fieldPath: metadata.name
+            - name: VAULT_CLUSTER_ADDR
+              value: "https://$(HOSTNAME).vault-internal:8201"
+            - name: HOME
+              value: "/home/vault"
+            - name: VAULT_LOG_LEVEL
+              value: "trace"
+
```
skabashnyuk, line 234: can you remove unnecessary spaces?

hack/vault-init.sh
sparkoo: it won't login after restart. Or do you see it logged in?

config/vault/openshift/deployment.yaml
sparkoo: these are generated by helm see `vault-generate-template.sh` script. I would not modify them manually. I can somehow update the script to remove empty lines, but I'm not sure it's worth it.

hack/vault-init.sh
skabashnyuk: I've missed the point where the code made logout. is there the one?

config/vault/openshift/deployment.yaml
skabashnyuk: ok

hack/vault-init.sh
skabashnyuk: ok

skabashnyuk (approved):
Looks good to me. what would be our next steps in the path to deploy and use vault on staging?

PR #80, opened 2022-02-23
Replace the delete webhook on SPIAccessToken with a finalizer.

metlos:
### What does this PR do?
`${TITLE}`

N/A

https://issues.redhat.com/browse/SVPI-76

### How to test this PR?
(the steps assume that we are still storing the token data as secrets)
1. Create an `SPIAccessTokenBinding`
1. Go through the OAuth flow so that it is injected
1. Make note of the linked access token
1. Check that there are 2 secrets in the namespace - 1 owned by the binding and 1 owned by the token
1. Delete the binding
1. Delete the token
1. Both secrets should disappear

### DIFF
--- a/.github/workflows/codecov.yaml
+++ b/.github/workflows/codecov.yaml
@@ -24,6 +24,8 @@ jobs:
         with:
           go-version: 1.16
       - name: Run tests
+        env:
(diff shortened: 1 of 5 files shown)


sparkoo:
do you have docker image to test ?

sparkoo (approved):

PR #82, opened 2022-02-28
Remove webhooks

metlos:
### What does this PR do?
This is a followup PR of #80 that removes the rest of the webhooks from the codebase.

https://issues.redhat.com/browse/SVPI-75, https://issues.redhat.com/browse/SVPI-76

### How to test this PR?
The only functional change is that we no longer check that a user making updates to the `SPIAccessTokenBinding` is also able to get `SPIAccessToken`s. All other functionality should be unaffected.


config/default/kustomization.yaml
```diff
@@ -25,7 +25,7 @@ resources:
 - ../namespace
 - ../crd
 - ../rbac
-- ../webhook
```
sparkoo, line 28: can we remove rather than just comment out?

config/k8s/kustomization.yaml
```diff
@@ -25,7 +25,7 @@ resources:
 - ../namespace
 - ../crd
 - ../rbac
-- ../webhook
```
sparkoo, line 28: remove

config/prepare/kustomization.yaml
```diff
@@ -15,7 +15,7 @@ resources:
 - ../namespace
 - ../crd
 - ../rbac
-- ../webhook
```
sparkoo, line 18: remove

main.go
```diff
@@ -148,23 +147,6 @@ func main() {
 		setupLog.Info("CRD controllers inactive")
 	}
 
-	if config.RunWebhooks() {
```
sparkoo, line 151: remove `RunWebhooks` function from config?

webhooks/spiaccesstoken_webhook.go
```diff
@@ -1,118 +0,0 @@
...
-	return &wh.Webhook{
-		Handler: w,
-	}
-}
-
-func (w *SPIAccessTokenWebhook) handleCreate(ctx context.Context, req wh.Request, t *api.SPIAccessToken) wh.Response {
-	changed := false
-
-	if t.Spec.RawTokenData != nil {
-		dataLocation, err := w.TokenStorage.Store(ctx, t, t.Spec.RawTokenData)
```
sparkoo, line 92: how is token stored now? `TokenStorage.Store` is not called anywhere.

sparkoo (changes requested):
Is it still WIP or is it ready to review? Especially part with storing the token is not clear to me.

webhooks/spiaccesstoken_webhook.go
sparkoo: stupid me, it's created from oauth service

config/default/kustomization.yaml
metlos: I've left it here because it was originally generated by operator-sdk and other (commented out, autogenerated) places still reference it.

sparkoo:
```
~/dev/spi-operator [metlos_remove-webhooks ✘]✖✹✭ λ SPIO_IMG=quay.io/mvala/spi-operator:metlos_remove-webhooks make deploy_minikube
Error: no matches for Id admissionregistration.k8s.io_v1_MutatingWebhookConfiguration|~X|mutating-webhook-configuration; failed to find unique target for patch admissionregistration.k8s.io_v1_MutatingWebhookConfiguration|mutating-webhook-configuration
error: no objects passed to apply
make: *** [Makefile:196: deploy_minikube] Error 1
```

sparkoo:
`go fmt` failed

sparkoo (approved):
Tested and works fine. Can you please clean commented code, as we discussed offline? I guess we must also update https://github.com/redhat-appstudio/service-provider-integration-oauth/pull/41

sparkoo:
works fine on minikube. I'll test on openshift

sparkoo (approved):
got it working on openshift with infra-deployments

PR #88, opened 2022-03-09
feat: Svpi61 vault token storage

sparkoo:
### What does this PR do?
Implements Vault token storage

spi-oauth counterpart https://github.com/redhat-appstudio/service-provider-integration-oauth/pull/55

https://issues.redhat.com/browse/SVPI-61

### How to test this PR?

 - `quay.io/mvala/spi-operator:svpi61-vaultTokenStorage`
 - `quay.io/mvala/spi-oauth:svpi61-vaultTokenStorage`

checkout and run `make deploy_minikube SPIO_IMG=quay.io/mvala/spi-operator:svpi61-vaultTokenStorage SPIS_IMG=quay.io/mvala/spi-oauth:svpi61-vaultTokenStorage`


pkg/spi-shared/config/config.go
```diff
@@ -50,6 +51,9 @@ type PersistedConfiguration struct {
 	// duration as string accepted by the time.ParseDuration function (e.g. "5m", "1h30m", "5s", etc.). The default
 	// is 1h (1 hour).
 	TokenLookupCacheTtl string `yaml:"tokenLookupCacheTtl"`
+
+	// VaultHost is url to Vault storage. Default `spi-vault` which is default spi Vault service name for kubernetes deployments.
```
sparkoo, line 55: fix default

pkg/spi-shared/config/config.go
```diff
@@ -71,6 +75,10 @@ type Configuration struct {
 
 	// TokenLookupCacheTtl is the time for which the lookup cache results are considered valid
 	TokenLookupCacheTtl time.Duration
+
+	VaultHost string
```
sparkoo, line 79: doc

pkg/spi-shared/config/config.go
```diff
@@ -71,6 +75,10 @@ type Configuration struct {
...
+
+	ServiceAccountTokenFilePath string
```
sparkoo, line 81: doc

integration_tests/suite_test.go
```diff
@@ -60,6 +62,7 @@ type IntegrationTest struct {
 }
 
 var ITest IntegrationTest
+var vaultTestCluster *vault.TestCluster
```
metlos, line 65: I'd like it better if this was part of the `ITest` global, so that we have just one place to find all the "support infrastructure" for the tests

integration_tests/suite_test.go
```diff
@@ -165,7 +168,8 @@ var _ = BeforeSuite(func() {
 	})
 	Expect(err).NotTo(HaveOccurred())
 
-	strg, err := tokenstorage.New(ITest.Client)
+	var strg tokenstorage.TokenStorage
+	vaultTestCluster, strg = tokenstorage.CreateTestVaultTokenStorage(&testing.T{})
```
metlos, line 172: Would `GinkgoT()` work instead of `&testing.T{}` here?

pkg/spi-shared/config/config.go
```diff
@@ -104,11 +115,21 @@ func (c PersistedConfiguration) inflate() (Configuration, error) {
...
+	} else {
+		conf.VaultHost = c.VaultHost
+	}
+
 	ttl, err := time.ParseDuration(ttlStr)
 	if err != nil {
 		return conf, err
+	}
 
+	if saTokenPath, ok := os.LookupEnv("SA_TOKEN_PATH"); ok {
```
metlos, line 129: This env var is not documented in the description of that config value. I think it would be worth mentioned there.

main.go
```diff
@@ -105,8 +105,7 @@ func main() {
 		os.Exit(1)
 	}
 
-	strg, err := tokenstorage.New(mgr.GetClient())
-
+	strg, err := tokenstorage.NewVaultStorage(mgr.GetClient(), "spi-controller-manager", cfg.VaultHost, cfg.ServiceAccountTokenFilePath)
```
metlos, line 108: Is it ok to hardcode the `role` to `spi-controller-manager` here?

pkg/spi-shared/tokenstorage/vault.go
```diff
@@ -0,0 +1,137 @@
...
+		return nil, fmt.Errorf("corrupted data in Vault at '%s'", path)
+	}
+
+	return parseToken(ctx, data)
+}
+
+func parseToken(ctx context.Context, data interface{}) (*api.Token, error) {
+	// We only get `interface{}` from Vault. In order to get `Token{}` struct,
+	// we first marshall interface into JSON, and then unmarshall it back to `Token{}` struct.
+	jsonData, err := json.Marshal(data)
```
metlos, line 111: what actual type is the `data` at runtime? it feels weird to convert it to JSON and then right back into a struct.

metlos (commented):
Overall 👍🏼 , just some minor comments below.

pkg/spi-shared/tokenstorage/vault.go
sparkoo: it's really `interface{}`. comes from this map https://github.com/hashicorp/vault/blob/main/api/secret.go#L26

sparkoo: if you have better idea how to convert it to our `Token`, I'm all in

integration_tests/suite_test.go
sparkoo: omg, I was looking at this, thinking it is not possible because of this guy https://cs.opensource.google/go/go/+/refs/tags/go1.18:src/testing/testing.go;l=766 only now I see that vault uses it's own interface created for this exact reason https://github.com/mitchellh/go-testing-interface/blob/master/testing.go#L18.

Thanks for making me looking at it again.

main.go
sparkoo: This is not kubernetes role. it is vault "equivalent to serviceaccount" and it is set up here https://github.com/redhat-appstudio/service-provider-integration-operator/pull/88/files#diff-862cfcfe83b93917c4a2f5b5e7f4c00fa4cb7af432e447d0b8eb442ea4a00d43R130. It is saying that role `spi-controller-manager` (which is just a name and can be anything) expects kubernetes serviceaccount `spi-controller-manager` from `spi-system` namespaces and gives this role `spi` vault permissions (defined here https://github.com/redhat-appstudio/service-provider-integration-operator/pull/88/files#diff-862cfcfe83b93917c4a2f5b5e7f4c00fa4cb7af432e447d0b8eb442ea4a00d43R130)

integration_tests/suite_test.go
sparkoo: done

pkg/spi-shared/config/config.go
sparkoo: done

pkg/spi-shared/tokenstorage/vault.go
metlos: well, the json marshaller is able to make sense of it and extract the data from it, so it cannot be just `interface{}` at runtime. I was just advocating to see if this cannot be reliably cast into at least `map[string]interface{}` or something from which we could extract the data, too, without the whole serde dance...

main.go
metlos: Ok, seems safe enough. If need be, we can make this configurable in the future to prevent clashes, but the name seems specific enough that I would not worry about it too much at this moment.

pkg/spi-shared/tokenstorage/vault.go
sparkoo: I've changed it to mapstructure decoder (comes with vault dependency), but I'm not sure about it. I can also get the fields individually, but it's not very nice code. One benefit I see with using json marshaller is, that it will follow Token struct tags. But I'm not sure it's worth as I agree it's kind of nasty hack.

WDYT?

metlos: I've tried to debug into the mapstructure and it is the same reflection-ridden madness as json decoder :)

what about just this (tested only with the test vault cluster, not the real one) (`json.Number` is just a typedef of `string` so the conversions in the below code are noops and the parsing into an actual `uint64` is inevitable):

I know that this code breaks as soon as we change the format of the data stored in Vault but that would most possibly be true for serialization solutions, too, if the data in vault is in the older format. At the same time this is going to be called a lot, so I think we could afford to be specific.

```go
func parseToken(data interface{}) (*api.Token, error) {
	token := &api.Token{}
	dataMap, ok := data.(map[string]interface{})
	if !ok {
		return nil, fmt.Errorf("unexpected data")
	}

	token.AccessToken = nonNilOrDefault(dataMap["access_token"], "").(string)
	token.TokenType = nonNilOrDefault(dataMap["token_type"], "").(string)
	token.RefreshToken = nonNilOrDefault(dataMap["refresh_token"], "").(string)
	expiryNum := nonNilOrDefault(dataMap["expiry"], json.Number("0")).(json.Number)

	expiry, err := strconv.ParseUint(expiryNum.String(), 10, 64)
	if err != nil {
		return nil, fmt.Errorf("invalid expiry value: %v", expiryNum)
	}
	token.Expiry = expiry

	return token, nil
}

func nonNilOrDefault(val interface{}, defaultValue interface{}) interface{} {
	if val == nil {
		val = defaultValue
	}

	return val
}
```

sparkoo: ok, next try :) It's basically your suggestion, except I very much don't like `interface{}`s in function args/return (doesn't lead to anything good from my experience), so I've wrote 2 separate functions for `string` and `uint64` (we can refactor to generics once we switch to 1.18 🚀). If you're ok with it, I'll write few tests.

Btw, can you please explain more why you don't like generic solution? I mean JSON or mapstructure one? Is it because it's hacky? Performance? Prone to errors?

metlos: For me it's performance.

sparkoo: ok. I agree with performance, as it is not using reflection (debatable how it is performance critical place though). I wouldn't be so sure with low maintenance cost. We need to always think to update this function when updating `Token`, which might lead to unnecessary debug sessions. Also, when we would like to enforce properties in `Token` json tags, the logic becomes tricky. But we can always refactor later.

metlos: We also need to think about how to deal with existing data when updating the `Token`. At that point I think the manual solution might even offer more flexibility. But I actually am not too worried about this. I don't want to block the PR because of this.

skabashnyuk (approved):

metlos (approved):

pkg/spi-shared/config/config_test.go
```diff
@@ -57,6 +57,7 @@ serviceProviders:
   clientId: "456"
   clientSecret: "54"
 baseUrl: blabol
+vaultHost: vaultTestHost
```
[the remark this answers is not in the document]
sparkoo: it's useful when running operator locally so you can connect to vault in cluster. So I would say NO to infra-deployments. I'll update operator's README.md.

PR #98, opened 2022-03-23
feat: Enhance error handling in the operator

metlos:
### What does this PR do?
This enhances the error handling in the `SPIAccessToken` controller.

Introduce Invalid and Error phases, error reason and error message to the SPIAccessToken status.
    
Enhance the github metadata provider with the ability to distinguish between token errors and other "environmental" errors.

Additionally, this removes the `AuthenticatedOAuthState` from the codebase of the operator. It is only used by the oauth service so it has no place in the operator codebase. Instead, the code in the operator was made more flexible to support any kind of datastructure encoded as the OAuth state (which will be used by the OAuth service).
N/A

This PR is a preparation for https://issues.redhat.com/browse/SVPI-62 - the endpoint for manual token upload

### How to test this PR?
1. Create an `SPIAccessToken` object with an invalid service provider URL.
2. The token phase should flip to "Invalid"

The other error states are currently not invokable using normal operation of the operator. This will only be possible once we have the manual upload endpoint.


skabashnyuk (approved):
works for me
Знімок екрана 2022-03-25 о 09 59 44

metlos:
Linter is passing locally but consistently times out in the CI. I will merge this PR but if the problem persists in other PRs we will have to take a look at that...

PR #109, opened 2022-04-05
Token metadata collection for Quay pt.1 (username & scopes)

mshaposhnik:
### What does this PR do?
Metadata colletion for quay tokens (mostly related to scopes and username)

https://issues.redhat.com/browse/SVPI-88 (partial)

### How to test this PR?

Configure quay, fetch a token. Metadata must be present:

```
tokenMetadata:
    lastRefreshTime: 1649331136
    scopes:
      - repo:read
      - repo:write
      - user:read
    serviceProviderState: eyJBY2Nlc3NpYmxlUmVwb3MiOnt9LCJSZW1vdGVVc2VybmFtZSI6Im1zaGFwb3NobmlrIn0=
    userId: ''
    username: $oauthtoken
```    

and state contains encoded data with real username:

```
{"AccessibleRepos":{},"RemoteUsername":"mshaposhnik"}
```


skabashnyuk (approved):
works for me
Знімок екрана 2022-04-05 о 17 24 00


I was able to pull/push images with the OAuth token. The single problem I met is the username. It is a real username however to log in with this token we need to use `$oauthtoken`

pkg/serviceprovider/quay/metadataprovider.go
```diff
@@ -0,0 +1,111 @@
...
+	}
+
+	metadata := token.Status.TokenMetadata
+	if metadata == nil {
+		metadata = &api.TokenMetadata{}
+		token.Status.TokenMetadata = metadata
+	}
+
+	scopes := make([]string, 0)
+	for _, p := range token.Spec.Permissions.Required {
```
metlos, line 79: You're missing `token.Spec.Permissions.AdditionalScopes` here.

There is a utility method in the `serviceprovider` package for this kind of thing:

Extract `quay.TranslateToScopes` into a simple function (not a struct method), call it `translateToQuayScopes` or something and use `serviceprovider.GetAllScopes(translateToQuayScopes, &token.Spec.Permissions)` instead of the loop here.

metlos (commented):
LGTM apart from the one comment below.

pkg/serviceprovider/quay/metadataprovider.go
mshaposhnik: ok, applied

PR #114, opened 2022-04-20
update the spi component to use PVC instead of volumeClaimTemplate

mbharatk:
### What does this PR do?

As PVC is created for the whole namespace and the pipelineRun is using subPath, the PVC was being full very soon, and to overcome this, as a workaround we used volumeClaimTemplate, which creates PVC for each pipelineRun. 

Now, there is a [PVC-cleaner](https://github.com/redhat-appstudio/pvc-cleaner) which will delete the subPath when a pipelineRun is deleted. 
This PR is to switch back to subPaths from volumeClaimTemplate.

https://issues.redhat.com/browse/PLNSRVCE-174

### DIFF
--- a/.tekton/pull_request.yaml
+++ b/.tekton/pull_request.yaml
@@ -21,10 +21,6 @@ spec:
   serviceAccountName: pipeline
   workspaces:
     - name: workspace
-      volumeClaimTemplate:
-        spec:
-          accessModes:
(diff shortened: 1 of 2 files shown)


skabashnyuk (approved):

mbharatk:
@skabashnyuk could you please merge this PR?

PR #133, opened 2022-05-24
Ensure that secret type kubernetes.io/dockerconfigjson is properly generated for the Quay image provider and can be used as Deployment's imagePullSecret

xbaran4:

### What does this PR do?
When a user specifies secret type as kubernetes.io/dockerconfigjson in binding, this PR ensures a proper secret is created that can be used as an imagepullsecret in podspec.

binding:
image
secret:
image
successfully pulled image:
image

fixes: https://issues.redhat.com/browse/SVPI-115

### How to test this PR?

1. deploy operator with changes; wait for all components to be running
` make deploy_minikube SPIO_IMG=quay.io/pbaran/spio:SVPI-115-DOCKERCONFIG`

2. create a token binding that has a secret type: kubernetes.io/dockerconfigjson and repo URL that specifies a private quay image that can be able pulled to a pod, such as: 
`kubectl apply -f samples/binding-dockerconfigjson.yaml`

3. go through the OAuth flow

4. check that dockerconfigjson secret was created and create a Pod that uses it as imagepullsecret:
`kubectl apply -f test-pod`
where test-pod looks like: 
```yaml
apiVersion: v1
kind: Pod
metadata:
  name: private-reg-pod
spec:
  containers:
  - name: private-reg-container
    image: <private_image>
  imagePullSecrets:
  - name: test-secret-dockerconfigjson
```
5. check that the pod has successfully pulled the image


controllers/access_token_mapper.go
```diff
@@ -49,7 +50,10 @@ func (at AccessTokenMapper) toSecretType(secretType corev1.SecretType) map[strin
 	case corev1.SecretTypeDockercfg:
 		ret[corev1.DockerConfigKey] = at.Token
 	case corev1.SecretTypeDockerConfigJson:
-		ret[corev1.DockerConfigJsonKey] = at.Token
+		ret[corev1.DockerConfigJsonKey] = fmt.Sprintf(`{"auths":{"%s":{"username":"%s","password":"%s"}}}`,
+			strings.TrimPrefix(at.ServiceProviderUrl, "https://"),
```
metlos, line 54: I'd rather parse the URL and just use the host part of it... We can't be 100% sure the scheme will always be https.

metlos (commented):
LGTM apart from the SP URL handling.

controllers/access_token_mapper.go
```diff
@@ -49,7 +50,14 @@ func (at AccessTokenMapper) toSecretType(secretType corev1.SecretType) map[strin
 	case corev1.SecretTypeDockercfg:
 		ret[corev1.DockerConfigKey] = at.Token
 	case corev1.SecretTypeDockerConfigJson:
-		ret[corev1.DockerConfigJsonKey] = at.Token
+		url := at.ServiceProviderUrl
+		if split := strings.Split(at.ServiceProviderUrl, "://"); len(split) > 1 {
```
sparkoo, line 54: I think @metlos meant to use `url.Parse` and then just `parsed.Host`. Something like here https://github.com/redhat-appstudio/service-provider-integration-operator/blob/main/pkg/serviceprovider/lookup.go#L50

controllers/access_token_mapper_test.go
```diff
@@ -74,6 +74,13 @@ func TestSecretTypeDefaultFields(t *testing.T) {
 		assert.Equal(t, `{"auths":{"spurl":{"username":"spusername","password":"token"}}}`, converted[corev1.DockerConfigJsonKey])
 	})
 
+	t.Run("dockerconfigjson-urlWithScheme", func(t *testing.T) {
+		newAt := at // copy to not affect other tests
+		newAt.ServiceProviderUrl = "http://quay.io/somepath"
+		converted := newAt.toSecretType(corev1.SecretTypeDockerConfigJson)
+		assert.Equal(t, `{"auths":{"quay.io/somepath":{"username":"spusername","password":"token"}}}`, converted[corev1.DockerConfigJsonKey])
```
metlos, line 81: This is actually incorrect imho, because the authentications seem to be stored per host, not per host+path.

metlos (changes requested):

skabashnyuk:
not working for me. I think because of https://github.com/redhat-appstudio/service-provider-integration-operator/pull/146

xbaran4:
> not working for me. I think because of #146
If you set only "r" permissions it should be working despite #146

sparkoo (approved):

metlos (approved):

PR #138, opened 2022-05-30
Use UBI as our base image

metlos:
### What does this PR do?
`$TITLE`.

Apart from the standardization on UBI across App Studio, this will also improve debugability by allowing shell access to the container.

https://issues.redhat.com/browse/SVPI-131

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -18,9 +18,19 @@ COPY pkg/ pkg/
 # Build
 RUN CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build -a -o manager main.go
 
-# Use distroless as minimal base image to package the manager binary
-# Refer to https://github.com/GoogleContainerTools/distroless for more details
-FROM gcr.io/distroless/static:nonroot
+# Compose the final image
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -18,9 +18,19 @@ COPY pkg/ pkg/
 # Build
 RUN CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build -a -o manager main.go
 
-# Use distroless as minimal base image to package the manager binary
-# Refer to https://github.com/GoogleContainerTools/distroless for more details
-FROM gcr.io/distroless/static:nonroot
+# Compose the final image
+FROM registry.access.redhat.com/ubi8/ubi-minimal:8.5-218
```
skabashnyuk, line 22: What about to use `registry.access.redhat.com/ubi8/ubi-minimal:8.6` or `registry.access.redhat.com/ubi8/ubi-minimal:8.5` ?

mshaposhnik (approved):

Dockerfile
metlos: We're now using the latest version.

PR #140, opened 2022-06-03
Split single Github graphql request into two

skabashnyuk:
### What does this PR do?
Since a single request might lead to some sort of timeouts or other errors it makes sense to separate it into two independent and merge results in runtime.

n/a

https://issues.redhat.com/browse/SVPI-133

### How to test this PR?
1. deploy `quay.io/skabashn/service-provider-integration-operator:SVPI-133_2022_06_06__16_03_37` operator image
2. Create SpiAccessTokenBinding 
3. Upload `Sahil Budhwar'` Github token.
4. SpiAccessTokeToken has to have metadata.
5. Binding should work and token should be injected as a secret

### DIFF
--- a/pkg/serviceprovider/github/queries.go
+++ b/pkg/serviceprovider/github/queries.go
@@ -16,8 +16,9 @@ package github
 
 import (
 	"context"
-	"sigs.k8s.io/controller-runtime/pkg/log"
 	"strconv"
+
+	"sigs.k8s.io/controller-runtime/pkg/log"
 
 	sperrors "github.com/redhat-appstudio/service-provider-integration-operator/pkg/errors"
 	"github.com/redhat-appstudio/service-provider-integration-operator/pkg/spi-shared/httptransport"
--- a/pkg/serviceprovider/github/queries.go
+++ b/pkg/serviceprovider/github/queries.go
@@ -16,6 +16,7 @@ package github
 
 import (
 	"context"
+	"strconv"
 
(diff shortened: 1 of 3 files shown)


skabashnyuk:
From my experiments with my token. I got identical lists with 1163 repositories.

metlos (approved):

pkg/serviceprovider/github/queries.go
```diff
@@ -91,7 +99,27 @@ type PageInfo struct {
...
+	lg := log.FromContext(ctx)
+	err := _fetchAllWithPages(ctx, client, graphql.NewRequest(request), r, func() PageInfo {
+		return r.Viewer.Repositories.PageInfo
+	}, func() {
+		for _, node := range r.Viewer.Repositories.Nodes {
+			state.AccessibleRepos[RepositoryUrl(node.Url)] = RepositoryRecord{ViewerPermission: ViewerPermission(node.ViewerPermission)}
+		}
+	})
+	//TODO how to do lg.Debug ?
+	lg.Info("_fetchAll state.AccessibleRepos.len=" + strconv.Itoa(len(state.AccessibleRepos)))
```
sparkoo, line 113: it will be printed for each call, so this will be printed twice in logs. Wouldn't it be better to log in `FetchAll` function, so it's printed just once?

sparkoo (approved):
code looks fine. However, I don't know github api and I havent tested it on real deployment

PR #142, opened 2022-06-06
Add Snyk provider

mshaposhnik:

### DIFF
--- a/pkg/serviceprovider/snyk/snyk.go
+++ b/pkg/serviceprovider/snyk/snyk.go
@@ -40,6 +40,9 @@ func newSnyk(factory *serviceprovider.Factory, _ string) (serviceprovider.Servic
 				tokenStorage: factory.TokenStorage,
 			},
 			MetadataCache: &cache,
+			RepoHostParser: serviceprovider.RepoHostParserFunc(func(repoUrl string) (string, error) {
+				return repoUrl, nil
+			}),
 		},
 		httpClient: factory.HttpClient,
 	}, nil
--- a/pkg/serviceprovider/snyk/snyk.go
+++ b/pkg/serviceprovider/snyk/snyk.go
@@ -93,6 +93,14 @@ func (g *Snyk) CheckRepositoryAccess(ctx context.Context, cl client.Client, acce
 	}, nil
 }
 
+func (g *Snyk) MapToken(_ context.Context, _ *api.SPIAccessTokenBinding, token *api.SPIAccessToken, tokenData *api.Token) (serviceprovider.AccessTokenMapper, error) {
+	return serviceprovider.DefaultMapToken(token, tokenData)
+}
+
+func (g *Snyk) Validate(ctx context.Context, _ serviceprovider.Validated) (serviceprovider.ValidationResult, error) {
+	return serviceprovider.ValidationResult{}, nil
+}
+
 type snykProbe struct{}
 
 var _ serviceprovider.Probe = (*snykProbe)(nil)
--- a/pkg/serviceprovider/snyk/snyk.go
+++ b/pkg/serviceprovider/snyk/snyk.go
@@ -1,0 +1,106 @@
+package snyk
+
+import (
+	"context"
+	"net/http"
+	"strings"
+
+	api "github.com/redhat-appstudio/service-provider-integration-operator/api/v1beta1"
+	"github.com/redhat-appstudio/service-provider-integration-operator/pkg/serviceprovider"
+	"github.com/redhat-appstudio/service-provider-integration-operator/pkg/spi-shared/config"
+	"k8s.io/client-go/rest"
+	"sigs.k8s.io/controller-runtime/pkg/client"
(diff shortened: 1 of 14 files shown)


pkg/serviceprovider/snyk/snyk.go
```diff
@@ -0,0 +1,117 @@
...
+		ErrorMessage:  "Access check for snyk.io is not implemented.",
+	}, nil
+}
+
+func (g *Snyk) MapToken(_ context.Context, _ *api.SPIAccessTokenBinding, token *api.SPIAccessToken, tokenData *api.Token) (serviceprovider.AccessTokenMapper, error) {
+	return serviceprovider.DefaultMapToken(token, tokenData)
+}
+
+func (g *Snyk) Validate(ctx context.Context, _ serviceprovider.Validated) (serviceprovider.ValidationResult, error) {
+	return serviceprovider.ValidationResult{}, nil
```
mshaposhnik, line 104: is there anything i can validate ?

pkg/serviceprovider/snyk/snyk.go
```diff
@@ -0,0 +1,117 @@
...
+		Configuration: factory.Configuration,
+		lookup: serviceprovider.GenericLookup{
+			ServiceProviderType: api.ServiceProviderTypeSnyk,
+			TokenFilter:         &tokenFilter{},
+			MetadataProvider: &metadataProvider{
+				httpClient:   factory.HttpClient,
+				tokenStorage: factory.TokenStorage,
+			},
+			MetadataCache: &cache,
+			RepoHostParser: serviceprovider.RepoHostParserFunc(func(repoUrl string) (string, error) {
```
mshaposhnik, line 43: @metlos  is that loopback ok there ?

pkg/spi-shared/config/config.go
```diff
@@ -198,3 +199,7 @@ func readFrom(rdr io.Reader) (PersistedConfiguration, error) {
 
 	return conf, err
 }
+
+func (c ServiceProviderConfiguration) OAuthConfigured() bool {
```
mshaposhnik, line 203: To be used in OAuth

pkg/serviceprovider/snyk/snyk.go
metlos: This is used to pick all tokens that belong to the same host (i.e. all github tokens in a namespace, all quay tokens, etc.) so if this is enough to do that then it is ok.

pkg/serviceprovider/snyk/snyk.go
```diff
@@ -0,0 +1,117 @@
...
+func (g *Snyk) GetType() api.ServiceProviderType {
+	return api.ServiceProviderTypeSnyk
+}
+
+func (g *Snyk) TranslateToScopes(_ api.Permission) []string {
+	return []string{}
+}
+
+func (g *Snyk) LookupToken(ctx context.Context, cl client.Client, binding *api.SPIAccessTokenBinding) (*api.SPIAccessToken, error) {
+	tokens, err := g.lookup.Lookup(ctx, cl, binding)
```
metlos, line 70: This doesn't really have to call the generic look up actually. If we ever are going to support one token per namespace.. You can just `cl.List()` the first token labeled with the snyk host and return that...

But I guess that's what generic token lookup does anyway so maybe it is good to leave this code like it is.

pkg/serviceprovider/snyk/snyk.go
metlos: nope, because you don't really support any permissions. Maybe it would even be good to make `SPIAccessToken.Permissions` and `SPIAccessTokenBinding.Permissions` optional so that people are not confused what to specify there for providers that don't support any of those.

pkg/serviceprovider/snyk/tokenfilter.go
```diff
@@ -0,0 +1,17 @@
...
+	api "github.com/redhat-appstudio/service-provider-integration-operator/api/v1beta1"
+	"github.com/redhat-appstudio/service-provider-integration-operator/pkg/serviceprovider"
+)
+
+type tokenFilter struct{}
+
+var _ serviceprovider.TokenFilter = (*tokenFilter)(nil)
+
+func (t *tokenFilter) Matches(_ context.Context, matchable serviceprovider.Matchable, token *api.SPIAccessToken) (bool, error) {
+	//TODO: implement token matching
```
metlos, line 15: We don't have anything to do the matching with, so I guess this really is the implementation. No need to leave the TODO here, maybe just an explanation why we're simply always matching.

metlos (commented):
As is, the code looks ok, but I was thinking maybe we could generalize this a little bit and have some kind of fallback host-based provider that would work with provider, for which we only allowed single token per namespace and host-based token matching (i.e. the same that you implemented here).

Do you think it would make sense or do you see anything in the your snyk implementation that would make it not work with this generic one (perhaps the fact that your matching both `snyk.io` and `api.snyk.io` in the `Probe.Examine` impl for snyk?).

PR #143, opened 2022-06-06
Use https protocol where it possible

skabashnyuk:
### What does this PR do?
Use HTTPS protocol where it is possible

n/a

After the https://github.com/redhat-appstudio/service-provider-integration-oauth/pull/92 OAuth service is use HTTPS only cookies. This pr aligns configuration with the requirement to use HTTPS where possible.

### How to test this PR?
n/a

### DIFF
--- a/bundle/manifests/service-provider-integration-operator.clusterserviceversion.yaml
+++ b/bundle/manifests/service-provider-integration-operator.clusterserviceversion.yaml
@@ -152,7 +152,7 @@ spec:
                 - /manager
                 env:
                 - name: SPI_URL
-                  value: http://service-provider-integration-api
+                  value: https://service-provider-integration-api
                 image: quay.io/redhat-appstudio/service-provider-integration-operator:next
                 livenessProbe:
                   httpGet:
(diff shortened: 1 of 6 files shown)


mshaposhnik (approved):

bundle/manifests/service-provider-integration-operator.clusterserviceversion.yaml
```diff
@@ -152,7 +152,7 @@ spec:
                 - /manager
                 env:
                 - name: SPI_URL
```
metlos, line 154: I think this is actually not used anymore.

config/manager/manager.yaml
```diff
@@ -25,7 +25,7 @@ spec:
         - --leader-elect
         env:
         - name: SPI_URL
```
metlos, line 27: I think this is actually not used anymore.

metlos (approved):
Looks ok. I'm fine if you don't want to remove the `SPI_URL`  in this PR, but I think it is safe to remove because it is no longer used by the code.

PR #144, opened 2022-06-08
Add Common username + token provider

mshaposhnik:
### What does this PR do?

As it is supposed that Snyk is not the last provider to support `username` + `token` only manual upload ability, it was deciced to do that provider universal, i..e support any URL-s and perform tokens matching by them. One can still provide attitional info during upload, but it will not be taken into account during matching. So the prinicple is `single token per namespace per host`.

[]
(https://issues.redhat.com/browse/SVPI-118)

### How to test this PR?
 - Build image;
 - Create SPIAccessTokenBinding for any URL;
 Example:
 ```
kind: SPIAccessTokenBinding
metadata:
  namespace: default
spec:
  repoUrl: https://snyk.io/
  secret:

 ```
 - Check binding created and grab linked SPIAccessToken name;
 - Perform usual manual token uploading, containing at least username + token;
 
 ```
curl -k -v -X POST -H "Content-Type: application/json" -H "Authorization: Bearer $(hack/get-default-sa-token.sh)" -d'{"username": "uy7c+spi_demo", "access_token": "4R28N79MT"}' "http://spi.$(minikube ip).nip.io/token/default/generated-spi-access-token-5sp5c"
 ```
 - Make sure token is `Ready`, secrets created, etc
 - Create another binding to the same URL. It should match to existing token immediately.
 - Create another binding with different URL. New SPIAccessToken with `AwaitingToken` state must be created.


pkg/serviceprovider/common/common.go
```diff
@@ -0,0 +1,124 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/log"
+)
+
+type Common struct {
+	Configuration config.Configuration
+	lookup        serviceprovider.GenericLookup
+	httpClient    rest.HTTPClient
+	repoUrl       string
+}
+
```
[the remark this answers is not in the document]
mshaposhnik: functions are described in the interface. for this given class - agree? deserves few words.

skabashnyuk:
I have this error 
Знімок екрана 2022-06-13 о 16 28 29

pkg/serviceprovider/common/tokenfilter.go
```diff
@@ -0,0 +1,31 @@
...
+	api "github.com/redhat-appstudio/service-provider-integration-operator/api/v1beta1"
+	"github.com/redhat-appstudio/service-provider-integration-operator/pkg/serviceprovider"
+)
+
+type tokenFilter struct{}
+
+var _ serviceprovider.TokenFilter = (*tokenFilter)(nil)
+
+func (t *tokenFilter) Matches(_ context.Context, _ serviceprovider.Matchable, _ *api.SPIAccessToken) (bool, error) {
+	// Matching is done via hostname in token, so does not need additional matching gere.
```
sparkoo, line 29:
```suggestion
	// Matching is done via hostname in token, so does not need additional matching here.
```

sparkoo (approved):

pkg/serviceprovider/common/common.go
```diff
@@ -0,0 +1,124 @@
...
+	"strings"
+
+	api "github.com/redhat-appstudio/service-provider-integration-operator/api/v1beta1"
+	"github.com/redhat-appstudio/service-provider-integration-operator/pkg/serviceprovider"
+	"github.com/redhat-appstudio/service-provider-integration-operator/pkg/spi-shared/config"
+	"k8s.io/client-go/rest"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/log"
+)
+
```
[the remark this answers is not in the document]
metlos: 👍🏼 This provider stores username+password pairs and resolves them by host. So maybe I'd call it `HostCredentialsServiceProvider` or something like that...

pkg/serviceprovider/common/common.go
```diff
@@ -0,0 +1,129 @@
...
+func newCommon(factory *serviceprovider.Factory, repoUrl string) (serviceprovider.ServiceProvider, error) {
+
+	cache := serviceprovider.NewMetadataCache(factory.KubernetesClient, &serviceprovider.NeverMetadataExpirationPolicy{})
+	return &Common{
+		Configuration: factory.Configuration,
+		lookup: serviceprovider.GenericLookup{
+			ServiceProviderType: api.ServiceProviderTypeCommon,
+			TokenFilter:         &tokenFilter{},
+			RepoHostParser: serviceprovider.RepoHostParserFunc(func(repoUrl string) (string, error) {
+				schemeIndex := strings.Index(repoUrl, "://")
```
metlos, line 53: Since this is just copied from Quay, I think it would make sense to actually extract this to a separate function, say `RepoHostFromSchemelessUrl` and use it here and in Quay to avoid repetition...

pkg/serviceprovider/common/metadataprovider.go
```diff
@@ -0,0 +1,55 @@
...
+	lg := log.FromContext(ctx, "tokenName", token.Name, "tokenNamespace", token.Namespace)
+
+	data, err := p.tokenStorage.Get(ctx, token)
+	if err != nil {
+		lg.Error(err, "failed to get the token metadata")
+		return nil, err
+	}
+
+	if data == nil {
+		return nil, err
```
metlos, line 43: IMHO `err` here is confusing. It is always nil, so what about `return nil, nil`?

skabashnyuk:
Works for me
Знімок екрана 2022-06-15 о 11 15 17
Знімок екрана 2022-06-15 о 11 30 48
I would say it make sense to rename 
```json
    "labels": {
      "spi.appstudio.redhat.com/service-provider-host": "snyk.io",
      "spi.appstudio.redhat.com/service-provider-type": "Common"
    },
```
to
```json
    "labels": {
      "spi.appstudio.redhat.com/service-provider-host": "snyk.io",
      "spi.appstudio.redhat.com/service-provider-type": "HostCredentialsProvider"
    },
```

wdyt?
One more thing
what about moving content of `hostcredentials` package to `serviceproviders`

README.md
```diff
@@ -251,7 +251,45 @@ The currently supported list of providers and token kinds is following:
...
+
+ - Determine the SPIAccessTokenName:
+ ```
+SPI_ACCESS_TOKEN=$(kubectl get spiaccesstokenbindings test-access-token-binding -o=jsonpath='{.status.linkedAccessTokenName}')
+```
+
+
+ - Perform manual push of the credential data: 
+```
+curl -k -v -X POST -H "Content-Type: application/json" -H "Authorization: Bearer $(hack/get-default-sa-token.sh)" -d'{"username": "userfoo", "access_token": "4R28N79MT"}' "http://<cluster-host-or-ip>/token/default/$SPI_ACCESS_TOKEN"
```
skabashnyuk, line 291: http->https

skabashnyuk (approved):

metlos (approved):
It took me a while to figure out why this provider can be used as a fallback and doesn't interfere with the probing of the repo URLs in other providers. I think a comment at the right place would make that much clearer. Otherwise I think this looks good!

pkg/serviceprovider/hostcredentials/hostcredentials.go
```diff
@@ -0,0 +1,121 @@
...
+// only one token for particular URL in the given namespace.
+type HostCredentialsProvider struct {
+	Configuration config.Configuration
+	lookup        serviceprovider.GenericLookup
+	httpClient    rest.HTTPClient
+	repoUrl       string
+}
+
+var Initializer = serviceprovider.Initializer{
+	Constructor: serviceprovider.ConstructorFunc(newHostCredentialsProvider),
```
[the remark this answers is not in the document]
mshaposhnik: Ok

PR #147, opened 2022-06-09
Migrate to google/go-github for Github API request

skabashnyuk:
### What does this PR do?
Uses google/go-github go library to get the list of user's repositories. Replaces graphql queries since they are not returning private repositories that belongs to organizations 

Знімок екрана 2022-06-13 о 11 42 21

 https://issues.redhat.com/browse/SVPI-140 

### How to test this PR?
1. Deploy `quay.io/skabashn/service-provider-integration-operator:SVPI-140_2022_06_13__23_13_00`
2.  Create CR to the private repository that belongs to the organization 
```
kind: SPIAccessTokenBinding
metadata:
  name: binding-write-repository
spec:
  permissions:
    required:
      - type: w
        area: repository
  repoUrl: https://github.com/redhat-appstudio-qe/private-quarkus-devfile-sample
  secret:
    name: token-write-repository

```
3.  Upload Personal access token or complete OAuth flow. Note: user has to have an access to this repository.
4. Binding has to be matched and token has to pass injection.


pkg/serviceprovider/github/githubclientbuilder.go
```diff
@@ -0,0 +1,28 @@
...
+}
+
+func (g *githubClientBuilder) createAuthenticatedGhClient(ctx context.Context, spiToken *api.SPIAccessToken) (*github.Client, error) {
+	token, tsErr := g.tokenStorage.Get(ctx, spiToken)
+	if tsErr != nil {
+		lg := log.FromContext(ctx)
+		lg.Error(tsErr, "failed to get token from storage for", "token", spiToken)
+		return nil, tsErr
+	}
+	ctx = context.WithValue(context.TODO(), oauth2.HTTPClient, g.httpClient)
```
sparkoo, line 25: I think we should use `ctx` instead of `TODO()`

skabashnyuk:
`logs.go`   - is logger sugaring can be a subject to move to a separate pr if needed.

pkg/serviceprovider/github/queries.go
```diff
@@ -16,123 +16,62 @@ package github
...
 			} `json:"nodes"`
 		} `json:"repositories"`
 	} `json:"viewer"`
 }
 
-func (r *AllAccessibleRepos) FetchAll(ctx context.Context, client *graphql.Client, accessToken string, state *TokenState) error {
+func (r *AllAccessibleRepos) FetchAll(ctx context.Context, githubClient *github.Client, accessToken string, state *TokenState) error {
+
+	lg := log.FromContext(ctx)
+	defer logs.TimeTrack(lg, time.Now(), "fetch all github repositories")
```
sparkoo, line 42: does timing here works as expected? Is `time.Now()` really evaluated at the beginning of the method? If yes, is it guaranteed by go specification?

pkg/logs/logs.go
```diff
@@ -0,0 +1,49 @@
...
+}
+
+// TimeTrack used to time any function
+// Example:
+//  {
+//    defer logs.TimeTrack(lg, time.Now(), "fetch all github repositories")
+//  }
+func TimeTrack(log logr.Logger, start time.Time, name string) {
+	elapsed := time.Since(start)
+	log.V(1).Info("Time took", "name", name, "time", elapsed)
```
sparkoo, line 48: this is really nice, but is it worth lvl 1? I guess we should have some consistent "guidelines" what levels to use for what.

sparkoo (commented):
This PR is adding 2 major things, that are not related together - github client and logging. I would rather have it separate next time, if possible. Fortunately, the PR is not that big.

Code looks fine, I have just few comments to the logging.

I haven't tested the changes in my deployment.

pkg/serviceprovider/github/queries.go
skabashnyuk: Yeh. It works. I've checked. Inspiration comes from https://coderwall.com/p/cp5fya/measuring-execution-time-in-go

pkg/logs/logs.go
skabashnyuk: AFAIK levels are just different levels of debugging depth. I have no clue why we need more than 1. 
Inspired by https://sdk.operatorframework.io/docs/building-operators/golang/references/logging/#a-simple-example

skabashnyuk: Would you like to have a constants or something like that?

pkg/serviceprovider/github/queries.go
sparkoo: Cool. I see it's in spec "Each time a "defer" statement executes, the function value and parameters to the call are [evaluated as usual](https://go.dev/ref/spec#Calls) and saved anew but the actual function is not invoked." https://go.dev/ref/spec#Defer_statements so it's safe to use.

sparkoo (approved):

mshaposhnik (approved):

PR #148, opened 2022-06-09
initial APIResourceSchema and APIExport for KCP

sparkoo:
### What does this PR do?
 - add `APIResourceSchema` for all our CRDs
 - add `APIExport` linking all `APIResourceSchema`s
 - add script to generate new version of `APIResourceSchema` on CRD change
 - make script runnable by `make manifests-kcp`, also added as dependency of `make manifests`

https://issues.redhat.com/browse/SVPI-91

### How to test this PR?

script requires `kubectl-kcp` to run properly. I've tried to write simple instructions how to install it.

 - run `make manifests-kcp`, you should see something like `no changes in KCP API`
 - do some change in one of CRDs, change some field name for example
 - run `make manifests-kcp` again. you should see updated `apiresourceschema_spi.yaml` file with your changes and current datetime as a name prefix. `apiexport_spi.yaml` must contain updated schemas with latest datetime.

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -57,7 +57,9 @@ jobs:
           echo "$HOME/bin" >> $GITHUB_PATH
       - name: Install kcp tooling
         run: |
-          git clone https://github.com/kcp-dev/kcp ~/kcp && cd ~/kcp && make install
+          git clone https://github.com/kcp-dev/kcp ~/kcp
+          cd ~/kcp
+          WHAT=./cmd/kubectl-kcp make install
       - name: Cache go modules
         id: cache-mod
         uses: actions/cache@v3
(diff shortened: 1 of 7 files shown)


.github/workflows/pr.yml
```diff
@@ -55,6 +55,11 @@ jobs:
           mkdir -p ~/bin
           cp ~/cache/operator-sdk-${OPERATOR_SDK_VERSION} ~/bin/operator-sdk
           echo "$HOME/bin" >> $GITHUB_PATH
+      - name: Install kcp tooling
+        run: |
+          git clone https://github.com/kcp-dev/kcp ~/kcp
+          cd ~/kcp
+          WHAT=./cmd/kubectl-kcp make install
```
skabashnyuk, line 62: how we will use this?

sparkoo: it's needed for "Check manifests" where we run `make manifests`

skabashnyuk (approved):
works for me
Знімок екрана 2022-06-10 о 11 02 40
Знімок екрана 2022-06-10 о 11 02 51

PR #151, opened 2022-06-13
Logger sugaring

skabashnyuk:
### What does this PR do?
Added a bit of logger sugar.
 - A separate method to initialize operator logger, global logger that can be reused from code
 - Setup logger in individual packages with Test_Main method
 - TimeTrack - method to measure method execution time.
 

Context https://github.com/redhat-appstudio/service-provider-integration-operator/pull/147#pullrequestreview-1004160666

Before 
Знімок екрана 2022-06-13 о 15 15 52

After

Знімок екрана 2022-06-13 о 15 14 35

https://issues.redhat.com/browse/SVPI-140

### How to test this PR?
Build and run test in some of the packages. `make test` for some reason has no affect.

### DIFF
--- a/pkg/logs/logs.go
+++ b/pkg/logs/logs.go
@@ -1,0 +1,53 @@
+//
+// Copyright (c) 2021 Red Hat, Inc.
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package logs
+
+import (
+	"flag"
+	"time"
+
+	"github.com/go-logr/logr"
+	"github.com/go-logr/zapr"
+	"go.uber.org/zap"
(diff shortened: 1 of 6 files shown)


pkg/logs/logs.go
```diff
@@ -0,0 +1,53 @@
...
+
+	"github.com/go-logr/logr"
+	"github.com/go-logr/zapr"
+	"go.uber.org/zap"
+	ctrl "sigs.k8s.io/controller-runtime"
+	crzap "sigs.k8s.io/controller-runtime/pkg/log/zap"
+)
+
+const (
+	DebugLvl = 1
```
sparkoo, line 29: does it make sense to think about more levels now?

sparkoo (approved):
cool. Do we have way how to configure log level in runtime?

sparkoo:
> AFAIK levels are just different levels of debugging depth. I have no clue why we need more than 1.
Inspired by https://sdk.operatorframework.io/docs/building-operators/golang/references/logging/#a-simple-example

I've assumed that level is here so we have more granularity. Looking at that example, does it mean `lg.Info(...)` is level 0 and that's what we will see in prod. And `lg.V(1).Info(...)` for development?

pkg/logs/logs.go
skabashnyuk: I think we need to collect some experience of the usage and then make some decision. I have no sense how to use more than 1 level now.

skabashnyuk:
>I've assumed that level is here so we have more granularity. Looking at that example, does it mean lg.Info(...) is level 0 and that's what we will see in prod. And lg.V(1).Info(...) for development?

Yes. Something like that.

PR #172, opened 2022-07-06
Enhanced logger configuration

skabashnyuk:
### What does this PR do?

- Adds the ability to configure a logger with command line and environment variables.
- Moved periodical non-error messages from INFO to DEBUG level
- Setup zap backend for `controller-runtime` ,`k8s.io/klog` , `hashicorp/go-hclog` 

```
Options:
  --metrics-bind-address METRICS-BIND-ADDRESS, -m METRICS-BIND-ADDRESS
                         The address the metric endpoint binds to. [default: :8080, env: METRICSADDR]
  --health-probe-bind-address HEALTH-PROBE-BIND-ADDRESS, -h HEALTH-PROBE-BIND-ADDRESS
                         The address the probe endpoint binds to. [default: :8081, env: PROBEADDR]
  --leader-elect, -l     Enable leader election for controller manager. Enabling this will ensure there is only one active controller manager. [default: false, env: ENABLELEADERELECTION]
  --config-file CONFIG-FILE, -c CONFIG-FILE
                         The location of the configuration file. [default: /etc/spi/config.yaml, env: CONFIGFILE]
  --vault-insecure-tls, -i
                         Whether is allowed or not insecure vault tls connection. [default: false, env: VAULTINSECURETLS]
  --zap-devel, -d        Development Mode defaults(encoder=consoleEncoder,logLevel=Debug,stackTraceLevel=Warn) Production Mode defaults(encoder=jsonEncoder,logLevel=Info,stackTraceLevel=Error) [default: false, env: ZAPDEVEL]
  --zap-encoder ZAP-ENCODER, -e ZAP-ENCODER
                         Zap log encoding (‘json’ or ‘console’) [env: ZAPENCODER]
  --zap-log-level ZAP-LOG-LEVEL, -v ZAP-LOG-LEVEL
                         Zap Level to configure the verbosity of logging [env: ZAPLOGLEVEL]
  --zap-stacktrace-level ZAP-STACKTRACE-LEVEL, -s ZAP-STACKTRACE-LEVEL
                         Zap Level at and above which stacktraces are captured [env: ZAPSTACKTRACELEVEL]
  --zap-time-encoding ZAP-TIME-ENCODING, -t ZAP-TIME-ENCODING
                         one of 'epoch', 'millis', 'nano', 'iso8601', 'rfc3339' or 'rfc3339nano' [default: rfc3339, env: ZAPTIMEENCODING]
  --help, -h             display this help and exit
```

prod mod
Знімок екрана 2022-07-06 о 16 25 48

dev mode
Знімок екрана 2022-07-06 о 18 29 36
command
```
kubectl set env deployment/spi-controller-manager ZAPDEVEL=true -n spi-system
```
### How to test this PR?
- Deploy `quay.io/skabashn/service-provider-integration-operator:newconfig_2022_07_12__10_12_04`
- Run `kubectl set env deployment/spi-controller-manager ZAPDEVEL=true -n spi-system` to enable dev mode.  Logger config `encoder=consoleEncoder,logLevel=Debug,stackTraceLevel=Warn`
- Run `kubectl set env deployment/spi-controller-manager ZAPDEVEL=false -n spi-system` to enable dev mode.  Logger config `encoder=jsonEncoder,logLevel=Info,stackTraceLevel=Error`

### DIFF
--- a/main.go
+++ b/main.go
@@ -62,6 +62,7 @@ type cliArgs struct {
 	ProbeAddr            string `arg:"-h, --health-probe-bind-address, env" default:":8081" help:"The address the probe endpoint binds to."`
 	EnableLeaderElection bool   `arg:"-l, --leader-elect, env" default:"false" help:"Enable leader election for controller manager. Enabling this will ensure there is only one active controller manager."`
 	ConfigFile           string `arg:"-c, --config-file, env" default:"/etc/spi/config.yaml" help:"The location of the configuration file."`
+	VaultInsecureTLS     bool   `arg:"-i, --vault-insecure-tls, env" default:"false" help:"Whether is allowed or not insecure vault tls connection."`
 	ZapDevel             bool   `arg:"-d, --zap-devel, env" default:"false" help:"Development Mode defaults(encoder=consoleEncoder,logLevel=Debug,stackTraceLevel=Warn) Production Mode defaults(encoder=jsonEncoder,logLevel=Info,stackTraceLevel=Error)"`
 	ZapEncoder           string `arg:"-e, --zap-encoder, env" default:"" help:"Zap log encoding (‘json’ or ‘console’)"`
 	ZapLogLevel          string `arg:"-v, --zap-log-level, env" default:"" help:"Zap Level to configure the verbosity of logging"`
@@ -101,7 +102,7 @@ func main() {
 		os.Exit(1)
 	}
 
-	strg, err := tokenstorage.NewVaultStorage("spi-controller-manager", cfg.VaultHost, cfg.ServiceAccountTokenFilePath, false)
+	strg, err := tokenstorage.NewVaultStorage("spi-controller-manager", cfg.VaultHost, cfg.ServiceAccountTokenFilePath, args.VaultInsecureTLS)
 	if err != nil {
 		setupLog.Error(err, "failed to initialize the token storage")
 		os.Exit(1)


main.go
```diff
@@ -95,13 +110,13 @@ func main() {
...
 
-	cfg, err := sharedConfig.LoadFrom(configFile)
+	cfg, err := sharedConfig.LoadFrom(args.ConfigFile)
 	if err != nil {
 		setupLog.Error(err, "Failed to load the configuration")
 		os.Exit(1)
 	}
 
-	strg, err := tokenstorage.NewVaultStorage("spi-controller-manager", cfg.VaultHost, cfg.ServiceAccountTokenFilePath, devmode)
+	strg, err := tokenstorage.NewVaultStorage("spi-controller-manager", cfg.VaultHost, cfg.ServiceAccountTokenFilePath, false)
```
metlos, line 119: Having the ability to configure insecure connection to vault comes in handy when running the operator locally and vault in in minikube for example.

pkg/logs/hclog.go
```diff
@@ -0,0 +1,184 @@
...
+		name: l.name,
+		zap:  l.zap,
+	}
+}
+
+func (l *HCLogAdapter) Log(level hclog.Level, msg string, args ...interface{}) {
+	var fields []zapcore.Field
+	for i := 0; i < len(args); i += 2 {
+		fields = append(fields, zap.Any(args[i].(string), args[i+1]))
+	}
```
metlos, line 60: Extract this into a function and maybe do some range checking to panic with some nicer message?

pkg/logs/logs.go
```diff
@@ -29,17 +35,43 @@ const (
...
-func InitLoggers(development bool, fs *flag.FlagSet) {
+func InitLoggers(development bool, encoder string, logLevel string, stackTraceLevel string, timeEncoding string) {
+
+	flagSet := flag.NewFlagSet("zap", flag.ContinueOnError)
+
+	opts := crzap.Options{ZapOpts: []zap.Option{zap.WithCaller(true), zap.AddCallerSkip(-1)}}
+	opts.BindFlags(flagSet)
+	_ = flagSet.Set("zap-devel", strconv.FormatBool(development))
+	if len(encoder) > 0 {
+		_ = flagSet.Set("zap-encoder", encoder)
```
metlos, line 52: Shouldn't we panic with some kind of output on `stderr` explaining the configuration error here?

pkg/serviceprovider/lookup.go
```diff
@@ -84,21 +86,21 @@ func (l GenericLookup) Lookup(ctx context.Context, cl client.Client, matchable M
 		return result, fmt.Errorf("failed to list the potentially matching tokens: %w", err)
 	}
 
-	lg.Info("lookup", "potential_matches", len(potentialMatches.Items))
+	lg.V(logs.DebugLvl).Info("lookup", "potential_matches", len(potentialMatches.Items))
```
metlos, line 89: I would much rather have a facade that I could use to say `logs.Debug(lg, msg, keysAndValues)` as opposed to the `lg.V(level).Info(...)` syntax, but I'm willing to surrender if you think otherwise :)

sparkoo: I also like more the slf4j way than this. But I'm not sure it's worth it. It will bring another dependency and it will become less go-like.

sparkoo:
What about bring this topic to other appstudio teams and get some agreement how logging should look like, from both developer and observer pov? If there is nothing yet about this topic, maybe we can propose some initial simple logging guidelines.

pkg/serviceprovider/lookup.go
metlos: Let me surrender here for now then :) Let's see how `lg.V(logs.DebugLevel).Info(...)` works for us..

metlos (approved):
I think this is a great step forward. There's only one remaining nitpick - I'd rename `logs.DebugLvl` to `logs.DebugLevel`.

sparkoo (approved):
maybe update README?

main.go
```diff
@@ -57,36 +57,37 @@ func init() {
 	//+kubebuilder:scaffold:scheme
 }
 
+type cliArgs struct {
+	MetricsAddr          string `arg:"-m, --metrics-bind-address, env" default:":8080" help:"The address the metric endpoint binds to."`
+	ProbeAddr            string `arg:"-h, --health-probe-bind-address, env" default:":8081" help:"The address the probe endpoint binds to."`
+	EnableLeaderElection bool   `arg:"-l, --leader-elect, env" default:"false" help:"Enable leader election for controller manager. Enabling this will ensure there is only one active controller manager."`
+	ConfigFile           string `arg:"-c, --config-file, env" default:"/etc/spi/config.yaml" help:"The location of the configuration file."`
+	VaultInsecureTLS     bool   `arg:"-i, --vault-insecure-tls, env" default:"false" help:"Whether is allowed or not insecure vault tls connection."`
```
sparkoo, line 65: I wouldn't specify `-i`, it's too specific config option to have shortcut like this.

PR #180, opened 2022-07-07
update status of spiaccesscheck when token lookup fails

sparkoo:
### What does this PR do?
Reports error to spiaccesscheck status when token lookup fails and don't reconcile CR again.

https://issues.redhat.com/browse/SVPI-161

### How to test this PR?

 1. manually upload the token with test/test credentials for quay.io token
```
curl -v \
  -H 'Content-Type: application/json' \
  -H "Authorization: bearer "$K8S_TOKEN \
  -d "{ \"access_token\": \"test\",\"username\": \"test\"}" \
  $OAUTH_URL'/token/'$SPI_NAMESPACE'/'$SPI_TOKEN_NAME
```
 2. create spiaccesscheck for quay.io repository
 3. you should see error in spiaccesscheck status that lookup failed. If repository is public, it should still be detected.


pkg/serviceprovider/quay/quay_test.go
```diff
@@ -505,13 +505,13 @@ func TestCheckRepositoryAccess(t *testing.T) {
...
+		assert.NoError(t, err)
 		assert.NotNil(t, status)
-		assert.False(t, status.Accessible)
+		assert.True(t, status.Accessible)
 		assert.Equal(t, api.SPIRepoTypeContainerRegistry, status.Type)
 		assert.Equal(t, api.ServiceProviderTypeQuay, status.ServiceProvider)
-		assert.Equal(t, api.SPIAccessCheckAccessibilityUnknown, status.Accessibility)
-		assert.Equal(t, api.SPIAccessCheckErrorUnknownError, status.ErrorReason)
+		assert.Equal(t, api.SPIAccessCheckAccessibilityPublic, status.Accessibility)
+		assert.Equal(t, api.SPIAccessCheckErrorTokenLookupFailed, status.ErrorReason)
```
metlos, line 514: I'm not sure how I would parse this as a user... Was there an error and "public" was left there as the default option or was the access determined to be public and the error was left there as an "artifact" of processing?

In either case I'd be confused whether this is a valid state...

sparkoo: fixed

pkg/serviceprovider/github/github.go
```diff
@@ -182,7 +182,11 @@ func (g *Github) CheckRepositoryAccess(ctx context.Context, cl client.Client, ac
 	tokens, lookupErr := g.lookup.Lookup(ctx, cl, accessCheck)
 	if lookupErr != nil {
 		lg.Error(lookupErr, "failed to lookup token for accesscheck", "accessCheck", accessCheck)
-		return status, fmt.Errorf("failed to lookip token for accesscheck %s/%s: %w", accessCheck.Namespace, accessCheck.Name, lookupErr)
+		if !publicRepo {
```
metlos, line 185: Btw why do we even continue here if we already determined the repository was public? Do we learn any new information by doing the token lookup?

pkg/serviceprovider/quay/quay.go
```diff
@@ -240,8 +243,10 @@ func (q *Quay) CheckRepositoryAccess(ctx context.Context, cl client.Client, acce
 		case http.StatusUnauthorized, http.StatusForbidden:
 			lg.Info("quay.io request unauthorized. Probably private repository for we don't have a token.")
```
metlos, line 244: "Probably a private repository for which we don't have a token."

metlos (commented):
LGTM apart from the two nitpicks..

pkg/serviceprovider/quay/quay.go
sparkoo: fixed

pkg/serviceprovider/github/github.go
sparkoo: idea was that we can find more information. But it's not actually needed. I've updated the code so it returns immediately at `if publicRepo`

metlos (approved):

PR #182, opened 2022-07-12
Defaults and environment variables as tags on config struct

metlos:
### What does this PR do?
Declare the default values and names of env vars that can override the configured values as tags on the PersistedConfiguration struct fields.

Slightly reshaped the Vault configuration in preparation for the future enhancements.

NA

### How to test this PR?
It should be possible to declare the newly defined environment variables on the operator pod to override the configuration options specified within the configuration file.

### DIFF
--- a/pkg/spi-shared/config/config.go
+++ b/pkg/spi-shared/config/config.go
@@ -18,6 +18,8 @@ import (
 	"io"
 	"io/ioutil"
 	"os"
+	"reflect"
+	"strings"
 	"time"
 
 	"gopkg.in/yaml.v3"
@@ -29,7 +31,6 @@ const (
 	ServiceProviderTypeGitHub          ServiceProviderType = "GitHub"
 	ServiceProviderTypeQuay            ServiceProviderType = "Quay"
 	ServiceProviderTypeHostCredentials ServiceProviderType = "HostCredentials"
-	DefaultVaultHost                   string              = "http://spi-vault:8200"
 )
 
 // PersistedConfiguration is the on-disk format of the configuration that references other files for shared secret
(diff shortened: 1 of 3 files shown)


sparkoo (commented):
do I understand correctly that env variables has higher priority than config file?

Looks like significant change in configuration and logic around configuration, can you please update README?

Will all current configurations in repo work? https://github.com/redhat-appstudio/service-provider-integration-operator/blob/main/config/openshift-example/config.yaml https://github.com/redhat-appstudio/service-provider-integration-operator/blob/main/config/k8s/config.yaml

Do we need to update configuration on staging env? If so, can you please create PR there and make sure that this is merged only when staging deployment is ready for it?

pkg/spi-shared/config/config.go
```diff
@@ -139,17 +156,68 @@ func (c PersistedConfiguration) inflate() (Configuration, error) {
...
+func (c *PersistedConfiguration) applyDefaults() {
+	applyTag(reflect.ValueOf(c), "default", false, func(val string) string {
+		return val
+	})
+}
+
+// applyTag is a helper function to PersistedConfiguration.apply* methods. It is assumed that the values are either
+// strings, arrays of strings or structs (which is true for PersistedConfiguration and its sub-structs). All other
+// types (like maps, ints, ...) are ignored.
+func applyTag(object reflect.Value, tagName string, overrides bool, valueFn func(tagValue string) string) {
```
sparkoo, line 182: isn't there some lib in our dependencies that can do something like this? It doesn't look very straightforward. If not, I'd like to see unit-tests for this, to catch as much weird corner-cases from the beginning.

metlos:
> lgtm, few questions:
> 
> do I understand correctly that env variables has higher priority than config file?
> 
Yes, so that you can override some of the configuration from the config file using the env vars in some kind of firefighting scenario.

> Looks like significant change in configuration and logic around configuration, can you please update README?
> 

Good point.

> Will all current configurations in repo work? https://github.com/redhat-appstudio/service-provider-integration-operator/blob/main/config/openshift-example/config.yaml https://github.com/redhat-appstudio/service-provider-integration-operator/blob/main/config/k8s/config.yaml
> 

Yes, they should both work. The only change was the move of the vault host to another place in the config file, but I believe we're always using the default value for that even in infra-deployments, right?

> Do we need to update configuration on staging env? If so, can you please create PR there and make sure that this is merged only when staging deployment is ready for it?

We shouldn't have to do it unless we set the vault host in those config files. As far as I looked we don't set that value in the config file.

pkg/spi-shared/config/config.go
metlos: Maybe we could use github.com/imdario/mergo (already in our indirect deps) or I looked at https://github.com/spf13/viper but none of those supports the definition using tags on the fields themselves. I personally like the proximity of the field and the definition of its default value and environment variable override in this approach as opposed to the code-based approach that we would have to adopt if we used either of those alternatives.

PR #185, opened 2022-07-13
Cleanup strategy #2

mshaposhnik:
Signed-off-by: Max Shaposhnik <mshaposh@redhat.com>


controllers/spiaccesstoken_controller.go
```diff
@@ -157,6 +157,14 @@ func (r *SPIAccessTokenReconciler) Reconcile(ctx context.Context, req ctrl.Reque
 		return ctrl.Result{}, nil
 	}
 
+	if at.Status.Phase == api.SPIAccessTokenPhaseAwaitingTokenData {
+		//if we passed finalizers, and token is in Awaiting, it means that it have no bindings referring to it and can be cleaned up
+		if err := r.Delete(ctx, &at); err != nil {
```
metlos, line 162: This is wrong, because it sets the deletion timestamp on the token. That's not what should happen IMHO. Why don't you do the same kind of check as is done in the "linkedBindingsFinalizer" here, before you try to delete?

metlos: Also, doing an "unguarded" delete like this most probably deletes the data from the token storage (because that is done by another finalizer).

mshaposhnik: >Why don't you do the same kind of check as is done in the "linkedBindingsFinalizer" here, before you try to delete?

I suppose that finalizaer was already execured (line 138), what's the point to perform same check again ?

mshaposhnik: @metlos  Is it looks better now ? (TBH i didn't fully have idea which data in storage we're talking about for `AwaitingData` tokens, can you explain that a little?)

metlos: yeah, I got confused there... there is no data in the token storage for tokens in awaiting phase... I just got worried that because you deleted unconditionally, all of the finalizers ran and one of those is the finalizer to delete the token data from storage. But for these tokens there is no data there...

controllers/spiaccesstoken_controller.go
```diff
@@ -157,6 +157,24 @@ func (r *SPIAccessTokenReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
+			}
+			if hasLinkedBindings {
+				return
+			}
+			//if token is in Awaiting, and no linked bindings present, it means that it have no bindings referring to it and can be cleaned up
+			if err := r.Delete(ctx, &at); err != nil {
+				lg.Error(err, "failed to cleanup the processed token")
+				return
+			}
+		}()
```
metlos, line 175: This means that if we successfully delete the token in this anonymous function, we continue with the reconciliation, right? Not sure this is the correct logic. IMHO there is no need to "hide" the errors from the rest of the reconciliation function like this. The token is in a valid phase and we're reconciling that phase. If we fail, the reconciliation should fail and be retried (which happens automatically if you return an error).

PR #203, opened 2022-07-20
vault audit log to stdout

sparkoo:
### What does this PR do?
fixes issue with vault audit log setup

this was actually never working on Openshift, but in previous commit, we've included `set -e` in script, so it started failing

### How to test this PR?

run `./hack/vault-init.sh` on vault instance on OpenShift


psturc (approved):
lgtm, verified on my openshift cluster

PR #210, opened 2022-07-27
Add timeout cleanup

mshaposhnik:
### What does this PR do?

Adds configurable TTL for tokens and bindings.

https://issues.redhat.com/browse/SVPI-65

### How to test this PR?

 - Use `mshaposh\spio:latest`
 - Configure deployment with some smaller timeouts for testing  (`--token-ttl`, `--binding-ttl` or `kubectl set env deployment/spi-controller-manager TOKENLIFETIMEDURATION=1m -n spi-system`)
 - Create objects and make sure they're deleted after timeout reached.

### DIFF
--- a/controllers/spiaccesstoken_controller.go
+++ b/controllers/spiaccesstoken_controller.go
@@ -172,13 +172,13 @@ func (r *SPIAccessTokenReconciler) Reconcile(ctx context.Context, req ctrl.Reque
 	if (tokenLifetime > r.Configuration.AccessTokenTtl.Seconds()) || (at.Status.Phase == api.SPIAccessTokenPhaseAwaitingTokenData && tokenLifetime > GracePeriodSeconds) {
 		hasLinkedBindings, err := hasLinkedBindings(ctx, &at, r.Client)
 		if err != nil {
-			lg.Error(err, "failed to check linked bindings for token", "token", at.ObjectMeta.Name, "error", err)
+			lg.Error(err, "failed to check linked bindings for token", "error", err)
 			return ctrl.Result{}, fmt.Errorf("failed to check linked bindings for token: %w", err)
 		}
 		if !hasLinkedBindings {
 			err = r.Delete(ctx, &at)
 			if err != nil {
-				lg.Error(err, "failed to cleanup obsolete token", "binding", at.ObjectMeta.Name, "error", err)
+				lg.Error(err, "failed to cleanup obsolete token", "error", err)
 				return ctrl.Result{}, fmt.Errorf("failed to cleanup token on reaching the lifetime or being unreferenced: %w", err)
 			}
 			lg.V(logs.DebugLevel).Info("token being deleted on reaching іts lifetime or being unreferenced with awaiting state", "token", at.ObjectMeta.Name, "tokenLifetime", tokenLifetime, "accesstokenttl", r.Configuration.AccessTokenTtl.Seconds())
--- a/controllers/spiaccesstoken_controller.go
+++ b/controllers/spiaccesstoken_controller.go
@@ -179,11 +179,6 @@ func (r *SPIAccessTokenReconciler) Reconcile(ctx context.Context, req ctrl.Reque
 			return ctrl.Result{}, fmt.Errorf("failed to validate the object: %w", err)
 		}
 		if !hasLinkedBindings {
-			err := r.TokenStorage.Delete(ctx, &at)
-			if err != nil {
-				lg.Error(err, "failed to cleanup token data on reaching the max lifetime", "binding", at.ObjectMeta.Name, "error", err)
-				return ctrl.Result{}, fmt.Errorf("failed to cleanup token data on reaching the max lifetime: %w", err)
-			}
 			err = r.Delete(ctx, &at)
 			if err != nil {
 				lg.Error(err, "failed to cleanup token on reaching the max lifetime", "binding", at.ObjectMeta.Name, "error", err)
--- a/controllers/spiaccesstoken_controller.go
+++ b/controllers/spiaccesstoken_controller.go
@@ -195,7 +195,7 @@ func (r *SPIAccessTokenReconciler) Reconcile(ctx context.Context, req ctrl.Reque
 		hasBindings = &hasLinkedBindings
 	}
 
-	// cleanup tokens which are in awaiting state and have their parent binging removed.
+	// cleanup tokens which are in awaiting state and have their parent binding removed.
 	// grace period taken in account to avoid race condition during the object creation process (i.e. when token is created but not yet linked)
 	if at.Status.Phase == api.SPIAccessTokenPhaseAwaitingTokenData && tokenLifetime > NoLinkingBindingGracePeriodSeconds {
 		var hasLinkedBindingsLocal bool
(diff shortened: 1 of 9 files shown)


controllers/spiaccesstoken_controller.go
```diff
@@ -166,15 +166,49 @@ func (r *SPIAccessTokenReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
+	var hasBindings *bool
+
+	// cleanup tokens by lifetime
+	if tokenLifetime > r.Configuration.AccessTokenTtl.Seconds() {
 		hasLinkedBindings, err := hasLinkedBindings(ctx, &at, r.Client)
 		if err != nil {
 			lg.Error(err, "failed to validate the object", "token", at.ObjectMeta.Name, "error", err)
 			return ctrl.Result{}, fmt.Errorf("failed to validate the object: %w", err)
 		}
 		if !hasLinkedBindings {
```
sparkoo, line 181: so what happens if token has some bindings? when it will be deleted? Shouldn't we schedule reconcile for this binding? https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/reconcile/reconcile.go#L33

metlos: I think this will just wait until finalizers give it a clear go when all the bindings are removed. This reconciler should react on the changes in the bindings, too..

So in another words, the `Delete()` will set the deletion timestamp and we will remove the finalizers once there are no more bindings linking to the token at which point it will finally be deleted...

controllers/spiaccesstokenbinding_controller.go
```diff
@@ -128,6 +131,17 @@ func (r *SPIAccessTokenBindingReconciler) Reconcile(ctx context.Context, req ctr
 		return ctrl.Result{}, nil
 	}
 
+	// cleanup bindings by lifetime
```
metlos, line 134: What about the idea of letting the user specify the ttl using a new property in the binding spec (and only use the configured default in that property is not specified)?

controllers/spiaccesstoken_controller.go
mshaposhnik: So, 
1) Is this code correct?
2) Should i explicitly call deletion from token storage, or it can be done by the finalizer?

controllers/spiaccesstokenbinding_controller.go
mshaposhnik: sounds cool..!

sparkoo: -1. it sounds like reasonable and good idea, but why make our lives harder? nobody requested that from us. why write, test and maintain code that nobody wants.

controllers/spiaccesstoken_controller.go
```diff
@@ -166,15 +166,49 @@ func (r *SPIAccessTokenReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
+			lg.V(logs.DebugLevel).Info("token being cleaned up on reaching the max lifetime", "binding", at.ObjectMeta.Name)
+			return ctrl.Result{}, nil
+		}
+		hasBindings = &hasLinkedBindings
+	}
+
+	// cleanup tokens which are in awaiting state and have their parent binging removed.
+	// grace period taken in account to avoid race condition during the object creation process (i.e. when token is created but not yet linked)
+	if at.Status.Phase == api.SPIAccessTokenPhaseAwaitingTokenData && tokenLifetime > NoLinkingBindingGracePeriodSeconds {
+		var hasLinkedBindingsLocal bool
```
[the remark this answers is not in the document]
mshaposhnik: since is a pointer type variable, the code not becomes more readable if try to use it directly...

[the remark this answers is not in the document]
mshaposhnik: refactored into single `if` block

controllers/spiaccesstoken_controller.go
```diff
@@ -166,21 +166,22 @@ func (r *SPIAccessTokenReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
-	if at.Status.Phase == api.SPIAccessTokenPhaseAwaitingTokenData && time.Since(at.CreationTimestamp.Time).Seconds() > NoLinkingBindingGracePeriodSeconds {
+	tokenLifetime := time.Since(at.CreationTimestamp.Time).Seconds()
+
+	// cleanup tokens by lifetime or on being unreferenced by any binding in the AwaitingToken state
+	if (tokenLifetime > r.Configuration.AccessTokenTtl.Seconds()) || (at.Status.Phase == api.SPIAccessTokenPhaseAwaitingTokenData && tokenLifetime > GracePeriodSeconds) {
 		hasLinkedBindings, err := hasLinkedBindings(ctx, &at, r.Client)
 		if err != nil {
-			lg.Error(err, "failed to validate the object", "token", at.ObjectMeta.Name, "error", err)
-			return ctrl.Result{}, fmt.Errorf("failed to validate the object: %w", err)
+			lg.Error(err, "failed to check linked bindings for token", "token", at.ObjectMeta.Name, "error", err)
```
metlos, line 175: no need to add `token` to the logger, because controller runtime already added the value for it to the logger in the context.

controllers/spiaccesstoken_controller.go
```diff
@@ -166,21 +166,22 @@ func (r *SPIAccessTokenReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
+			return ctrl.Result{}, fmt.Errorf("failed to check linked bindings for token: %w", err)
 		}
 		if !hasLinkedBindings {
-			//if token is in Awaiting, and no linked bindings present, it means that it have no bindings referring to it and can be cleaned up
-			if err := r.Delete(ctx, &at); err != nil {
-				lg.Error(err, "failed to cleanup the processed token", "token", at.ObjectMeta.Name, "error", err)
-				return ctrl.Result{}, fmt.Errorf("failed to cleanup the processed token: %w", err)
+			err = r.Delete(ctx, &at)
+			if err != nil {
+				lg.Error(err, "failed to cleanup obsolete token", "binding", at.ObjectMeta.Name, "error", err)
```
metlos, line 181: `binding` should be `token` but is unnecessary anyway, because the logger already has the reconciled token in its values.

controllers/spiaccesstokenbinding_controller.go
```diff
@@ -128,6 +131,18 @@ func (r *SPIAccessTokenBindingReconciler) Reconcile(ctx context.Context, req ctr
...
+	bindingLifetime := time.Since(binding.CreationTimestamp.Time).Seconds()
+	if bindingLifetime > r.Configuration.AccessTokenBindingTtl.Seconds() {
+		err := r.Client.Delete(ctx, &binding)
+		if err != nil {
+			lg.Error(err, "failed to cleanup binding on reaching the max lifetime", "binding", binding.ObjectMeta.Name, "error", err)
```
metlos, line 139: no need to specify the `binding` in the logger, it is already done by the controller runtime.

metlos (approved):
LGTM apart from some minor logging improvement suggestions below.

controllers/spiaccesstokenbinding_controller.go
mshaposhnik: ok for all

PR #212, opened 2022-08-01
Failure to perform docker login no longer fails the whole quay token lookup

metlos:
### What does this PR do?
`$TITLE`

This is a simple fix of the problem in the referenced issue. We should maybe think about introducing the validity checks of tokens in a more formal fashion in the workflow, but this fix would be required even with them, IMHO. So let's here first fix the issue and think about the validity checks at some later point in time.

https://issues.redhat.com/browse/SVPI-176

### How to test this PR?
See the repro steps in https://issues.redhat.com/browse/SVPI-176 . Instead of entering the irrecoverable error state, the re-created binding simply enters awaiting token data as expected.

### DIFF
--- a/pkg/serviceprovider/quay/dockerlogin.go
+++ b/pkg/serviceprovider/quay/dockerlogin.go
@@ -22,25 +22,27 @@ import (
 	"io"
 	"net/http"
 
+	"github.com/redhat-appstudio/service-provider-integration-operator/pkg/logs"
+
 	"github.com/golang-jwt/jwt/v4"
 	"github.com/pkg/errors"
 	"sigs.k8s.io/controller-runtime/pkg/log"
 )
 
-var unsuccessfulLoginError = errors.New("login did not succeed")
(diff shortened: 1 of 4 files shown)


pkg/serviceprovider/quay/dockerlogin.go
```diff
@@ -49,54 +51,54 @@ func DockerLogin(ctx context.Context, cl *http.Client, repository string, userna
...
 		}()
 
 		bytes, err := io.ReadAll(res.Body)
 		if err != nil {
-			lg.Error(err, "failed to read the body of response", "statusCode", res.StatusCode)
+			debugLog.Error(err, "failed to read the body of response", "statusCode", res.StatusCode)
 			return "", errors.WithMessage(err, "failed to read the body of non-200 response to login attempt")
 		}
 
 		msg := string(bytes)
```
sparkoo, line 71: I guess we no longer need this `msg := string(bytes)`

sparkoo (approved):

PR #217, opened 2022-08-04
Metrics monitoring configuration

skabashnyuk:


config/prometheus/manager_metrics_service.yaml
```diff
@@ -7,9 +7,8 @@ metadata:
   namespace: system
 spec:
   ports:
-  - name: https
```
metlos, line 10: Is there a way of publishing the metrics over https?

PR #218, opened 2022-08-05
feat: enable the quay and github service providers configuration as o…

xbaran4:

### What does this PR do?
Allows users to create Quay or GitHub spiaccesstokenbinding without having Quay or GitHub service providers configured in oauth configuration secret. In this way, the binding is linked to an spiaccesstoken that is compatible only with manual token upload (and no OAuth flow).

fixes https://issues.redhat.com/browse/SVPI-155

### How to test this PR?

1. Deploy operator with image: quay.io/pbaran/spio:SVPI-155
2. Delete Quay and GitHub service providers from oauth configuration secret (restart operator if needed to reload config)
3. Create binding `kubectl apply -f samples/spiaccesstokenbinding-2.yaml` and check that it is linked to spiaccesstoken with service-provider-type: Quay
4. Create binding `kubectl apply -f samples/spiaccesstokenbinding-3.yaml` and check that it is linked to spiaccesstoken with service-provider-type: GitHub


metlos (commented):
What we're being asked here to do, IMHO, is to support manual token upload even without explicit configuration of the service providers in our configuration file.

There is nothing optional about these providers, it's just that they can function in this limited mode without any additional configuration.

IMHO, we should maybe enhance the approach taken with "KnownInitializers" and categorize these by whether they support manual upload without config or not by e.g. having the initializer itself report this fact.

IMHO, that would be cleaner than having two places (KnownInitializers and your new function) that would know about all our implementations..

metlos (approved):
I think this is a step in the right direction because it provides us with a mechanism to pick and choose the providers we want to activate. We might refine the algorithm or activation conditions in the future, but I think this is good for today's requirements.

PR #222, opened 2022-08-09
Simplified token matching

skabashnyuk:
### What does this PR do?
It is changing the way of SPIAccessToken filtering when we are searching it for a SPIAccessTokenBinding. 
This pr makes it identical to 'hostcredential' provider. When tokens are matched by the service provider and that is all. No permission checks are made. 

Знімок екрана 2022-08-09 о 17 07 31

- https://issues.redhat.com/browse/SVPI-184

### How to test this PR?
 - deploy `quay.io/skabashn/service-provider-integration-operator:match2_2022_08_11__16_03_45`
 - create binding for read `https://github.com/redhat-appstudio/service-provider-integration-operator`
 - Complete OAuth flow.
 - Make sure token is injected


mshaposhnik:
Generally OK. Would be great to work with namings a bit more.

pkg/serviceprovider/tokenfilter.go
```diff
@@ -33,3 +37,36 @@ var _ TokenFilter = (TokenFilterFunc)(nil)
...
+// MatchAllTokenFilter is a TokenFilter that match any token
+var MatchAllTokenFilter TokenFilter = TokenFilterFunc(func(ctx context.Context, binding Matchable, token *api.SPIAccessToken) (bool, error) {
+	debugLog := log.FromContext(ctx).V(logs.DebugLevel)
+	debugLog.Info("Unconditional token match", "token", token)
+	return true, nil
+})
+
+// FallBackTokenFilter is a TokenFilter that filters token with MainTokenFilter if Condition returns true or FallBackFilter if
+// Condition returns false.
+type FallBackTokenFilter struct {
```
mshaposhnik, line 50: Don't muck like `Fallback` inside `Fallback`... What if we call outer filter `ConditionalTokenFilter` ?

mshaposhnik (approved):

pkg/spi-shared/config/config.go
```diff
@@ -87,6 +87,9 @@ type Configuration struct {
 
 	// AccessTokenBindingTtl is time after that AccessTokenBinding will be deleted.
 	AccessTokenBindingTtl time.Duration
+
+	// An indicator that is telling to match all tokens of the given provider in the concrete namespace
```
metlos, line 91: I would maybe call this `MatchAnyTokenPolicy`.

main.go
```diff
@@ -73,6 +73,7 @@ type cliArgs struct {
 	VaultApproleSecretIdFilePath   string                       `arg:"--vault-secretid-filepath, env" default:"/etc/spi/secret_id" help:"Used with Vault approle authentication. Filepath with secret_id."`
 	VaultKubernetesSATokenFilePath string                       `arg:"--vault-k8s-sa-token-filepath, env" help:"Used with Vault kubernetes authentication. Filepath to kubernetes ServiceAccount token. When empty, Vault configuration uses default k8s path. No need to set when running in k8s deployment, useful mostly for local development."`
 	VaultKubernetesRole            string                       `arg:"--vault-k8s-role, env" default:"spi-controller-manager" help:"Used with Vault kubernetes authentication. Vault authentication role set for k8s ServiceAccount."`
+	MatchAllTokenAlgorithm         bool                         `arg:"--match-all-token-algorithm, env" default:"true" help:"An indicator that is telling to match all tokens of the given provider in the concrete namespace."`
```
metlos, line 76: What about `--token-match-policy` with the default `any` and possible values `any` and `exact`?

sparkoo (approved):

metlos (approved):

PR #230, opened 2022-08-18
Initial setup for monitoring

skabashnyuk:
### What does this PR do?
The goal is to restore the default metrics protection mechanism using kube-rbac-proxy that initially comes from operator-sdk templates.  See https://github.com/redhat-appstudio/service-provider-integration-operator/commit/df23e228#diff-7263fc341f1b59afdc8599e1f570cdd180c28bf8cca71ecee1e58b58131202c5
Later these metrics would be collected on staging using independently  installed prometheus with https://github.com/redhat-appstudio/infra-deployments/pull/563

Since we use complex postprocessing login that is generating actual customization https://github.com/redhat-appstudio/service-provider-integration-operator/blob/main/hack/replace_placeholders_and_deploy.sh and here 
https://github.com/redhat-appstudio/service-provider-integration-operator/blob/main/config/minikube/kustomization.yaml#L18 it is almost impossible to use the original patch approach since the number of containers would be different in different cases.

n/a

https://issues.redhat.com/browse/SVPI-182

### How to test this PR?
n/a
There is no simple k8s or openshift way to test that now.


config/manager/manager.yaml
```diff
@@ -70,6 +72,17 @@ spec:
             name: vault-approle
             readOnly: true
             subPath: role_id
+      - name: kube-rbac-proxy
+        image: gcr.io/kubebuilder/kube-rbac-proxy:v0.8.0
+        args:
+          - "--secure-listen-address=0.0.0.0:8443"
+          - "--upstream=http://127.0.0.1:8080/"
+          - "--logtostderr=true"
+          - "--v=10"
```
metlos, line 81: How verbose is this?

config/manager/manager.yaml
```diff
@@ -70,6 +72,17 @@ spec:
...
+        ports:
+          - containerPort: 8443
+            protocol: TCP
+            name: https
```
bamachrn, line 85: we have a servce for this port export right?

config/manager/manager.yaml
metlos: K, let's keep it for a little while, but let's revisit this in the next release or something..

config/manager/manager.yaml
bamachrn: ok

PR #235, opened 2022-08-26
add new makefile target to deploy to openshift kcp workload with Route

sparkoo:
### What does this PR do?
 - new kustomize overlay `kcp_openshift` that includes the Route to oauth service
 - new makefile target `make deploy_kcp_openshift`, which deploys ^
 - add helper scripts to create and apply workload cluster to kcp, including one that sync Routes

https://issues.redhat.com/browse/SVPI-138

### How to test this PR?

 1. deploy to KCP with `make deploy_kcp_openshift`
 2. you must see the Route created with some URL generated in it
 3. you can enable debug logging for oauth service, so you see the requests in logs
 4. `curl` the route, with some path. You must see logs in oauth service with incoming requests

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -220,7 +220,7 @@ deploy_kcp: ensure-tmp manifests kustomize
 deploy_kcp_openshift: ensure-tmp manifests kustomize
 	if [ -z ${VAULT_HOST} ]; then echo "VAULT_HOST must be set"; exit 1; fi
 	$(eval KCP_WORKSPACE?=$(shell kubectl kcp workspace . --short))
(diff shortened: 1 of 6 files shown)


mshaposhnik (approved):

metlos (commented):
What's the difference between `deploy_kcp` and `deploy_kcp_openshift`? Their definitions are identical, afaict, so is this just for "namespacing" reasons where you can keep multiple differently configured "rendered" deployments in `.tmp`?

sparkoo:
@skabashnyuk @metlos 

`deploy_openshift` - deploys CRDs
`deploy_kcp` - does not deploy CRDs, instead of that deploys APIResourceSchema + APIExport + APIBinding
`deploy_kcp_openshift` - same as `deploy_kcp` + Route

PR #237, opened 2022-08-30
Explicitly declare VAULTKUBERNETESROLE for oauth service and operator

skabashnyuk:
### What does this PR do?
After this pr https://github.com/redhat-appstudio/service-provider-integration-oauth/commit/00dfaf188d0952a331d0248c06a19f2cd868fdd3#diff-2873f79a86c0d8b3335cd7731b0ecf7dd4301eb19a82ef7a1cba7589b5252261L63
OAuth service started to use the default https://github.com/redhat-appstudio/service-provider-integration-operator/blob/main/pkg/spi-shared/tokenstorage/vault.go#L77
which is not correct because of https://github.com/redhat-appstudio/service-provider-integration-operator/blob/main/hack/vault-init.sh#L139 

Знімок екрана 2022-08-31 о 09 09 53
Знімок екрана 2022-08-31 о 09 09 59
Знімок екрана 2022-08-31 о 09 20 01
Знімок екрана 2022-08-31 о 09 20 10

OAuth service failed to start because of 

```
{"level":"error","ts":"2022-08-30T11:06:18.817Z","caller":"spi-oauth/main.go:103","msg":"failed to create token storage interface","error":"error while authenticating: unable to log in to auth method: unable to log in with Kubernetes auth: Error making API request.\n\nURL: PUT https://vault.192.168.49.2.nip.io/v1/auth/kubernetes/login\nCode: 500. Errors:\n\n* service account name not authorized","stacktrace":"main.main\n\t/spi-oauth/main.go:103\nruntime.main\n\t/usr/local/go/src/runtime/proc.go:250"}
Stream closed EOF for spi-system/spi-oauth-service-858f8cd57b-rldqs (oauth)
```
### How to test this PR?
1. Deploy on minikube `quay.io/skabashn/service-provider-integration-operator:VAULTKUBERNETESROLE_2022_08_31__08_59_21`  and `quay.io/skabashn/service-provider-integration-oauth:main_2022_08_31__09_14_15`
2. operator and OAuth service should be able to start


metlos:
I checked all the original properties and the VAULTKUBERNETESROLE is the only "important" one that differs. The only other difference is that operator originally used the "iso8601" time encoding and the oauth service used "rfc3339" by default. The new common default is "iso8601".

PR #249, opened 2022-09-07
Always update the upload URL in the SPIAccessToken status

metlos:
### What does this PR do?
Make sure to regenerate the upload URL just in case of BASE_URL changes.

none

### How to test this PR?
1. `make install deploy_minikube SPIO_IMG=quay.io/lkrejci/service-provider-integration-operator:upload-url-fix`
2. `kubectl apply -f samples/binding-read-repo.yaml`
3. Check the upload URL: `kubectl get spiaccesstokenbinding binding-read-repository -ojsonpath='{.status.uploadUrl}`
5. `kubectl -n spi-system patch configmap/spi-shared-environment-config --type=json -p='[{"op": "replace", "path": "/data/BASEURL", "value": "https://now.its.there"}]'`
6. `kubectl -n spi-system rollout restart deployment/spi-controller-manager`
3. Check the upload URL: `kubectl get spiaccesstokenbinding binding-read-repository -ojsonpath='{.status.uploadUrl}`


integration_tests/spiaccesstokencontroller_test.go
```diff
@@ -75,6 +77,85 @@ var _ = Describe("Create without token data", func() {
...
+
+	BeforeEach(func() {
+		ITest.TestServiceProvider.Reset()
+		origBaseUrl = ITest.OperatorConfiguration.BaseUrl
+		ITest.OperatorConfiguration.BaseUrl = "https://initial.base.url"
+		ITest.TestServiceProvider.GetOauthEndpointImpl = func() string {
+			return ITest.OperatorConfiguration.BaseUrl + "/test/oauth"
+		}
+
+		createdToken = &api.SPIAccessToken{
```
mshaposhnik, line 104: Is that test stable (keeping in mind that tokens w/out linked bindings may eventually dissappear due to cleaner)

metlos: The testsuite sets the TTL of bindings and tokens to 10s. So it imho doesn't make a difference. I would actually say that we should by default set the TTLs to a much larger value to lower the probability of the TTL intervention and only set it to short TTLs in the dedicated tests.

sparkoo (approved):

integration_tests/spiaccesstokencontroller_test.go
mshaposhnik: I am talking about cleaner of `Awaiting...` state tokens w/out linked bindings. They're did not have TTL. It probably works only due to some 2s grace period now.

metlos: I made the deletion grace period configurable and set it to 10s in the integration test setup.

PR #253, opened 2022-09-08
Replace owner reference between SPIAccessBinding and secrets with finalizer

mshaposhnik:
Signed-off-by: Max Shaposhnyk <mshaposh@redhat.com>

### What does this PR do?
As the owner references are completely ignored in KCP, it must be replaced with finalizer.

https://issues.redhat.com/browse/SVPI-192

### How to test this PR?

It does not change behavior, only the implementation way.
Build own image, or use `mshaposh/spio:latest`
Create binding, inject the token, then remove binding. Secret must be deleted as well.

Short demo:

https://user-images.githubusercontent.com/1651062/190981655-3700081b-247f-414c-837e-18b90b8aa4e4.mp4

https://user-images.githubusercontent.com/1651062/191333792-45a019cc-c950-4ddd-b832-214d26a35ea6.mp4


sparkoo (commented):
do we still use owner references in single-cluster installations?

do you have built images to test?

controllers/spiaccesstokenbinding_controller.go
```diff
@@ -260,7 +286,7 @@ func (r *SPIAccessTokenBindingReconciler) Reconcile(ctx context.Context, req ctr
 	// now that we set up the binding correctly, we need to clean up the potentially dangling secret (that might contain
 	// stale data if the data of the token disappeared from the token)
 	if binding.Status.Phase == api.SPIAccessTokenBindingPhaseAwaitingTokenData {
-		if err := r.deleteSyncedSecret(ctx, existingSyncedSecretName, binding.Namespace); err != nil {
+		if err := deleteSyncedSecret(ctx, r.Client, existingSyncedSecretName, binding.Namespace); err != nil {
 			lg.Error(err, "failed to delete the stale synced object")
 			// note that we don't actually set any error on the binding itself, because it no longer references the
 			// secret. The secret will get cleaned up once the binding is deleted because of the owner reference.
```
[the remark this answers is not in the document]
mshaposhnik: no, as for now binding must indicate an error IMO. thanks

mshaposhnik:
Due to some unclear situation with secrets on KCP service, PR cannot be fully validated on KCP, but it is checked that finalizer  is called on `SPIAccessBinding` deletion, so we can assume it should work after secrets issue will be fixed (https://issues.redhat.com/browse/SVPI-203 )

sparkoo (approved):

controllers/spiaccesstokenbinding_controller.go
```diff
@@ -81,9 +85,13 @@ type SPIAccessTokenBindingReconciler struct {
...
 func (r *SPIAccessTokenBindingReconciler) SetupWithManager(mgr ctrl.Manager) error {
 	r.syncer = sync.New(mgr.GetClient())
+	r.finalizers = finalizer.NewFinalizers()
+	if err := r.finalizers.Register(linkedSecretsFinalizerName, &linkedSecretsFinalizer{client: r.Client}); err != nil {
+		return fmt.Errorf("failed to register the linked secrets finalizer: %w", err)
+	}
+
 	err := ctrl.NewControllerManagedBy(mgr).
 		For(&api.SPIAccessTokenBinding{}).
-		Owns(&corev1.Secret{}).
```
sparkoo, line 86: as we discussed, please check the behavior after this change. I think we want to keep watching the secrets so we reconcile them back if there is some external change to them.

mshaposhnik: Good point. I've added the secrets watch.

controllers/spiaccesstokenbinding_controller.go
```diff
@@ -81,35 +85,47 @@ type SPIAccessTokenBindingReconciler struct {
...
+		})).
+		Watches(&source.Kind{Type: &corev1.Secret{}}, handler.EnqueueRequestsFromMapFunc(func(o client.Object) []reconcile.Request {
+			kcpWorkspace := logicalcluster.From(o)
+			ctx := context.TODO()
+			if !kcpWorkspace.Empty() {
+				ctx = logicalcluster.WithCluster(ctx, kcpWorkspace)
+			}
+
+			requests, err := r.filteredBindingsAsRequests(ctx, kcpWorkspace.String(), o.GetNamespace(), func(binding api.SPIAccessTokenBinding) bool {
+				// return binding.Status.SyncedObjectRef == toObjectRef(o)  //i wanted that. But version and kind of o is empty by some unknown reason
```
metlos, line 119: This a handler is for Secrets... so the groupId is "", the version is "v1" and the kind is "Secret".

mshaposhnik: I was logged that...no kind or version, only name is there (idfk why)

metlos: yes, but you're watching the secrets, so that handler is receiving the secrets that you asked for, right? So even though `o` doesn't have the kind and version, you can be sure it is the secrets, because that's what that handler function is configured to handle.

controllers/spiaccesstokenbinding_controller.go
```diff
@@ -139,6 +177,23 @@ func (r *SPIAccessTokenBindingReconciler) Reconcile(ctx context.Context, req ctr
 	lg = lg.WithValues("linked_to", binding.Status.LinkedAccessTokenName,
 		"phase_at_reconcile_start", binding.Status.Phase)
 
+	finalizationResult, err := r.finalizers.Finalize(ctx, &binding)
+	if err != nil {
+		// if the finalization fails, the finalizer stays in place, and so we don't want any repeated attempts until
+		// we get another reconciliation due to cluster state change
+		return ctrl.Result{Requeue: false}, fmt.Errorf("failed to finalize: %w", err)
```
sparkoo, line 184: you don't have to explicitly set `Requeue: false`, but nothing wrong with that...

sparkoo (approved):

controllers/spiaccesstokenbinding_controller.go
```diff
@@ -85,35 +89,46 @@ type SPIAccessTokenBindingReconciler struct {
...
+		})).
+		Watches(&source.Kind{Type: &corev1.Secret{}}, handler.EnqueueRequestsFromMapFunc(func(o client.Object) []reconcile.Request {
+			kcpWorkspace := logicalcluster.From(o)
+			ctx := context.TODO()
+			if !kcpWorkspace.Empty() {
+				ctx = logicalcluster.WithCluster(ctx, kcpWorkspace)
+			}
+
+			requests, err := r.filteredBindingsAsRequests(ctx, kcpWorkspace.String(), o.GetNamespace(), func(binding api.SPIAccessTokenBinding) bool {
+				return binding.Status.SyncedObjectRef.Kind == "Secret" && binding.Status.SyncedObjectRef.Name == toObjectRef(o).Name
```
metlos, line 123: `o.GetName()` should do the same trick as `toObjectRef(o).Name` I think.

metlos (approved):
One minor nitpick but otherwise LGTM.

PR #256, opened 2022-09-09
Do not use image short names

lcarva:
Image references that do not include the registry are ambiguous and can potentially cause security issues. See:
https://www.redhat.com/en/blog/be-careful-when-pulling-images-short-name

This PR changes image references from `hashicorp/vault:1.10.3` to `docker.io/hashicorp/vault:1.10.3`.


sparkoo (changes requested):
deployments are generated from [vault-helm](https://github.com/hashicorp/vault-helm) by [script](https://github.com/redhat-appstudio/service-provider-integration-operator/blob/main/hack/vault-generate-template.sh). We should inclue image to [hack/vault-helm-values.yaml](https://github.com/redhat-appstudio/service-provider-integration-operator/blob/main/hack/vault-helm-values.yaml) and regenerate these deployments. See [values.yaml](https://github.com/hashicorp/vault-helm/blob/main/values.yaml), I think you want to add `server.image.repository`.

sparkoo (approved):

PR #281, opened 2022-10-04
Fixed default metrics binding address

skabashnyuk:
### What does this PR do?
- Fixed default metrics binding address
- Minor table fixup

n/a

- align docs with the code.

### How to test this PR?
n/a


README.md
```diff
@@ -108,7 +108,7 @@ and oauth service to have the same value for them):
...
 | --config-file                                         | CONFIGFILE                     | /etc/spi/config.yaml  | The location of the configuration file.                                                                                                                                                                                            |
-| --metrics-bind-address                                | METRICSADDR                    | :8080                 | The address the metric endpoint binds to.                                                                                                                                                                                          |
+| --metrics-bind-address                                | METRICSADDR                    | 127.0.0.1:8080        | The address the metric endpoint binds to.                                                                                                                                                                                          |
```
metlos, line 111: While this truly is the default, I'm not sure it is the default that is observable externally because of the rbac-proxy. Maybe we should add a note here that while this is the default from the operator binary point of view, the metrics are still available externally through the authorized endpoint provided by rbac-proxy?

metlos (approved):

PR #287, opened 2022-10-05
Add ability to customize reset() function of TestServiceProvider

xbaran4:

### What does this PR do?
Adds the ability to provide a CustomizeReset() function to TestServiceProvider which is executed at the end of Reset() function.

An improvement upon #280

### How to test this PR?

`make itest`


mshaposhnik (approved):

integration_tests/serviceprovider_mock.go
```diff
@@ -36,6 +36,7 @@ type TestServiceProvider struct {
 	CheckRepositoryAccessImpl func(context.Context, client.Client, *api.SPIAccessCheck) (*api.SPIAccessCheckStatus, error)
 	MapTokenImpl              func(context.Context, *api.SPIAccessTokenBinding, *api.SPIAccessToken, *api.Token) (serviceprovider.AccessTokenMapper, error)
 	ValidateImpl              func(context.Context, serviceprovider.Validated) (serviceprovider.ValidationResult, error)
+	CustomizeReset            func()
```
metlos, line 39: Just from the purity point of view, I would make this accept the instance of the `TestServiceProvider` that is being customized so that the impl doesn't have to assume that the instance is available through a global variable.

I.e.
```go
CustomizeReset func(*TestServiceProvider)
```

metlos (approved):

PR #297, opened 2022-10-10
Fix tekton task

skabashnyuk:
### What does this PR do?
 - created a separate docker image for oauth service since DEVHAS does not support multi-target builds.
 - created separate TaskRun for oauth and controller component
 - synced  configuration with https://github.com/redhat-appstudio/jvm-build-service/tree/main/.tekton repository

n/a

https://issues.redhat.com/browse/SVPI-224 

### How to test this PR?
n/a

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -17,7 +17,7 @@ COPY controllers/ controllers/
 COPY pkg/ pkg/
 
 # Build
-RUN CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build -o bin/ -a ./cmd/...
+RUN CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build -o bin/ -a ./cmd/operator/operator.go
 
 # Compose the final image of spi-operator.
 # !!! This must be last one, because we want simple `docker build .` to build the operator image.
--- a/oauth.Dockerfile
+++ b/oauth.Dockerfile
@@ -18,7 +18,7 @@ COPY oauth/ oauth/
 COPY pkg/ pkg/
 
 # Build
-RUN CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build -o bin/ -a ./cmd/...
+RUN CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build -o bin/ -a ./cmd/oauth/oauth.go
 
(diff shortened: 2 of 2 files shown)


oauth.Dockerfile
```diff
@@ -0,0 +1,45 @@
...
+# cache deps before building and copying source so that we don't need to re-download as much
+# and so that source changes don't invalidate our downloaded layer
+RUN go mod download
+
+COPY static/ static/
+
+# Copy the go source
+COPY cmd/ cmd/
+COPY api/ api/
+COPY controllers/ controllers/
```
sparkoo, line 17: can it utilize docker cache from operator build? If not, I think we can remove `controllers` here and from original `Dockerfile` remove `static` and `oauth`.

sparkoo (commented):
I think it will need couple of fixes in makefile `docker-build` and `docker-push` targets.

oauth.Dockerfile
```diff
@@ -0,0 +1,44 @@
...
+COPY static/ static/
+
+# Copy the go source
+COPY cmd/oauth cmd/oauth
+COPY api/ api/
+COPY oauth/ oauth/
+COPY pkg/ pkg/
+
+# Build
+RUN CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build -o bin/ -a ./cmd/...
```
sparkoo, line 21: we can build just `oauth.go`

Dockerfile
```diff
@@ -11,44 +9,19 @@ COPY go.sum go.sum
...
 # Copy the go source
-COPY cmd/ cmd/
+COPY cmd/operator cmd/operator
 COPY api/ api/
 COPY controllers/ controllers/
-COPY oauth/ oauth/
 COPY pkg/ pkg/
 
 # Build
 RUN CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build -o bin/ -a ./cmd/...
```
sparkoo, line 20: build just `operator.go`

sparkoo (approved):
just nitpick to build only exact go binary in dockerfile. I see you've already copied just oauth or operator folder, so technically it would make no difference. I just think it would be a bit nicer, but up to you.

PR #302, opened 2022-10-13
custom user oauth configuration

sparkoo:
### What does this PR do?
makes it possible to configure oauth service using secret in same namespace as SPIAccessTokenBinding

https://issues.redhat.com/browse/SVPI-205

### How to test this PR?

images:
```
quay.io/mvala/service-provider-integration-operator:svpi205-userOauth
quay.io/mvala/service-provider-integration-oauth:svpi205-userOauth
```

!!! make sure token you're using to do oauth flow has list and get secret perms in token's namespace

1. deploy with some nonsense oauth configuration for one of service providers
2. create spiaccesstokenbinding and try to do oauth flow and see it fail
3. create secret with oauth configuration in same namespace as a token (see `samples/sp-secret.yaml`)
4. try to do oauth flow again, should pick oauth config from the secret


mshaposhnik (approved):

README.md
```diff
@@ -424,3 +424,25 @@ or accessibility checks must be expected. A new CR instance should be used to re
 Currently, the file retrievals are limited to GitHub repositories only, and files size up to 2 Megabytes.
 
 
+## User OAuth configuration
+
+In situations, when OAuth configurations set for SPI does not fit user's use case (like on-prem installations), one may define it's own OAuth service provider configuration in the Kubernetes secret:
```
metlos, line 429: grammar nazi: "In situations when the OAuth configuration of SPI does not fit the user's use case (like on-prem installations), one may define their own OAuth service provider configuration using a Kubernetes secret:"

oauth/oauth_config.go
```diff
@@ -0,0 +1,119 @@
...
+	"github.com/redhat-appstudio/service-provider-integration-operator/pkg/spi-shared/oauthstate"
+	"golang.org/x/oauth2"
+	corev1 "k8s.io/api/core/v1"
+	kuberrors "k8s.io/apimachinery/pkg/api/errors"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/log"
+)
+
+const (
+	oauthCfgSecretLabel             = "service-provider-integration/service-provider-config"
```
metlos, line 32: We use a different naming convention for labels, see: https://github.com/redhat-appstudio/service-provider-integration-operator/blob/89f1d64fb36772c5694e0ff0645206befd847275/controllers/labels.go#L17

Also the expected value of this label is the type of the service provider, which to me is a little bit confusing given the name of the label.

What about calling this just simply `spi.appstudio.redhat.com/service-provider-type`?

oauth/oauth_config.go
```diff
@@ -0,0 +1,119 @@
...
+		RedirectURL: c.redirectUrl(),
+	}
+
+	found, oauthCfgSecret, findErr := c.findOauthConfigSecret(ctx, info)
+	if findErr != nil {
+		return nil, findErr
+	}
+
+	if found {
+		if createOauthCfgErr := createConfigFromSecret(oauthCfgSecret, oauthCfg); createOauthCfgErr == nil {
```
metlos, line 60: maybe `initializeConfigFromSecret` would describe better what this is doing..

oauth/oauth_config.go
```diff
@@ -0,0 +1,119 @@
...
+
+func (c *commonController) findOauthConfigSecret(ctx context.Context, info *oauthstate.OAuthInfo) (bool, *corev1.Secret, error) {
+	lg := log.FromContext(ctx).WithValues("oauthInfo", info)
+
+	secrets := &corev1.SecretList{}
+	if listErr := c.K8sClient.List(ctx, secrets, client.InNamespace(info.TokenNamespace), client.MatchingLabels{
+		oauthCfgSecretLabel: string(c.Config.ServiceProviderType),
+	}); listErr != nil {
+		if kuberrors.IsForbidden(listErr) {
+			lg.Info("user is not able to read secrets")
```
metlos, line 82: "... to list or get secrets"

oauth/oauth_config_test.go
```diff
@@ -0,0 +1,480 @@
...
+		assert.Nil(t, oauthCfg)
+	})
+}
+
+type mockTracker struct {
+	listImpl func(gvr schema.GroupVersionResource, gvk schema.GroupVersionKind, ns string) (runtime.Object, error)
+}
+
+func (t *mockTracker) Add(obj runtime.Object) error {
+	//TODO implement me
```
metlos, line 449: These are not real TODOs, IMHO... We just don't need these methods yet in any of the tests, right?

README.md
sparkoo: fixed

oauth/oauth_config.go
sparkoo: changed to `v1beta1.SPIAccessTokenLinkLabel`

oauth/oauth_config.go
sparkoo: fixed

oauth/oauth_config_test.go
sparkoo: yes

metlos (approved):
This is a great first step in the right direction and I'm glad we're already identifying some potential improvements.

PR #310, opened 2022-10-18
measure token metadata fetching in providers

metlos:
### What does this PR do?
This introduces the metrics capturing the overall time it takes to fetch the token metadata in different service providers.

This is done separately for github and quay, because each of them uses a different technique for fetching the metadata. Github fetches all the available metadata for all repositories upfront, while Quay impl fetches the metadata for each requested repository separately for each token.

Therefore, it doesn't make sense to have a single metric that would measure both, because these two approaches are incomparable.

https://issues.redhat.com/browse/SVPI-103

### How to test this PR?
1. Run SPI with `quay.io/lkrejci/service-provider-integration-operator:svpi-103` and `quay.io/lkrejci/service-provider-integration-oauth:svpi-103` docker images
2. configure the OAuth apps in Github and Quay
3. Create bindings to github and quay.
4. Provide token data to those bindings either by manual upload or through oauth flow.
5. `kubectl -n spi-system exec deployment/spi-controller-manager -c manager -t -- curl --silent localhost:8080/metrics | grep redhat` should list `redhat_appstudio_spi_github_token_metadata_fetch_seconds_bucket` metrics and `redhat_appstudio_spi_quay_single_repo_metadata_fetch_seconds_bucket` metrics.


sparkoo (approved):
code looks fine, but it feels like there must be a library to help with this, instead of write all manually

PR #312, opened 2022-10-19
add GitLab as supported service provider

xbaran4:
### What does this PR do?
Adds GitLab as a new supported service provider.

Some caveats:
-  `read_user` scope is needed by default to fetch user metadata with token
- ServiceProviderState, used to figure out to which repositories the token has access to, will be implemented in its own PR https://issues.redhat.com/browse/SVPI-255
- improved ServiceProvider probing will be implemented in it's own PR after support for multiple service provider user oauth configurations (https://issues.redhat.com/browse/SVPI-249) https://issues.redhat.com/browse/SVPI-256

Implements https://issues.redhat.com/browse/SVPI-234

### How to test this PR?

Using OAuth flow: 
1. Deploy SPI with images: `quay.io/pbaran/spis:GITLAB` for oauth service and `quay.io/pbaran/spio:GITLAB` for operator
2. Create a private repository on some instance of GitLab (you can use https://gitlab.com)
3. Create an authorized app in your user settings, like in the picture below, where the redirect URI is your OAuth service URL + /gitlab/callback
image
4. Copy the client ID (a.k.a. application ID), Secret, and create a new GitLab entry in spi config like the example below:
```yaml
serviceProviders:
- type: GitHub
  clientId: "123"
  clientSecret: "42"
- type: GitLab
  clientId: "your client id..."
  clientSecret: "your client secret..."
  baseUrl: "https://gitlab.com"
```
5. Make sure that OAuth service and operator have loaded the new config.
6. Create a binding such as `samples/binding-write-repo-gitlab.yaml` where you replace the repoUrl with your own.
7. Copy OAuth URL from binding, login, and go through the OAuth flow.
8. SPIAccessToken should then look something like this: 
image

Using manual upload: The PR should work with GitLab's personal access tokens (PAT) by uploading the token on OAuth service: 
1. Same as 1. and 2. step in OAuth flow.
2. Create PAT in GitLab.
image
3. Create binding same as in OAuth flow step 6.
4. Copy the token and upload it through token upload URL in binding.
5. SpiAccessToken should look the same as in OAuth flow step 8.

### DIFF
--- a/api/v1beta1/spiaccesstoken_types.go
+++ b/api/v1beta1/spiaccesstoken_types.go
@@ -126,6 +126,7 @@ const (
 	PermissionAreaRepositoryMetadata PermissionArea = "repositoryMetadata"
 	PermissionAreaWebhooks           PermissionArea = "webhooks"
 	PermissionAreaUser               PermissionArea = "user"
+	PermissionAreaRegistry           PermissionArea = "registry"
 )
 
 // SPIAccessTokenStatus defines the observed state of SPIAccessToken
--- a/api/v1beta1/spiaccesstoken_types.go
+++ b/api/v1beta1/spiaccesstoken_types.go
@@ -82,6 +82,7 @@ type ServiceProviderType string
 const (
 	ServiceProviderTypeGitHub          ServiceProviderType = "GitHub"
 	ServiceProviderTypeQuay            ServiceProviderType = "Quay"
+	ServiceProviderTypeGitLab          ServiceProviderType = "GitLab"
 	ServiceProviderTypeHostCredentials ServiceProviderType = "HostCredentials"
 )
 
--- a/pkg/serviceprovider/gitlab/gitlab.go
+++ b/pkg/serviceprovider/gitlab/gitlab.go
@@ -2,9 +2,13 @@ package gitlab
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"net/http"
 	"strings"
+
+	"github.com/xanzy/go-gitlab"
+	"sigs.k8s.io/controller-runtime/pkg/log"
 
 	api "github.com/redhat-appstudio/service-provider-integration-operator/api/v1beta1"
 	opconfig "github.com/redhat-appstudio/service-provider-integration-operator/pkg/config"
@@ -14,6 +18,13 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
+var unsupportedScopeError = errors.New("unsupported scope for GitLab")
+var unsupportedAreaError = errors.New("unsupported permission area for GitLab")
+var userWritePermissionsNotSupportedError = errors.New("user write permissions are not supported by GitLab")
+
+// Temp
+var notGitlabUrlError = errors.New("not a gitlab repository url")
+
 var _ serviceprovider.ServiceProvider = (*Gitlab)(nil)
 
 type Gitlab struct {
@@ -22,22 +33,31 @@ type Gitlab struct {
 	metadataProvider *metadataProvider
 	httpClient       rest.HTTPClient
 	tokenStorage     tokenstorage.TokenStorage
+	glClientBuilder  gitlabClientBuilder
 	baseUrl          string
 }
 
 var _ serviceprovider.ConstructorFunc = newGitlab
 
-func newGitlab(factory *serviceprovider.Factory, baseUrl string) (serviceprovider.ServiceProvider, error) {
+var Initializer = serviceprovider.Initializer{
+	Probe:                        gitlabProbe{},
+	Constructor:                  serviceprovider.ConstructorFunc(newGitlab),
+	SupportsManualUploadOnlyMode: true,
+}
 
-	// in Quay, we invalidate the individual cached repository records, because we're filling up the cache repo-by-repo
-	// therefore the metadata as a whole never gets refreshed.
+func newGitlab(factory *serviceprovider.Factory, baseUrl string) (serviceprovider.ServiceProvider, error) {
 	cache := serviceprovider.NewMetadataCache(factory.KubernetesClient, &serviceprovider.NeverMetadataExpirationPolicy{})
-	mp := &metadataProvider{
-		tokenStorage:     factory.TokenStorage,
-		httpClient:       factory.HttpClient,
-		kubernetesClient: factory.KubernetesClient,
-		ttl:              factory.Configuration.TokenLookupCacheTtl,
+	glClientBuilder := gitlabClientBuilder{
+		httpClient:   factory.HttpClient,
+		tokenStorage: factory.TokenStorage,
 	}
+	mp := &metadataProvider{
+		tokenStorage:    factory.TokenStorage,
+		httpClient:      factory.HttpClient,
+		glClientBuilder: glClientBuilder,
+		baseUrl:         baseUrl,
+	}
+
 	return &Gitlab{
 		Configuration: factory.Configuration,
 		lookup: serviceprovider.GenericLookup{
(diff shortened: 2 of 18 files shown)


api/v1beta1/spiaccesstoken_types.go
```diff
@@ -125,6 +126,7 @@ const (
 	PermissionAreaRepositoryMetadata PermissionArea = "repositoryMetadata"
 	PermissionAreaWebhooks           PermissionArea = "webhooks"
 	PermissionAreaUser               PermissionArea = "user"
+	PermissionAreaRegistry           PermissionArea = "registry"
```
metlos, line 129: Is `registry` an image registry?

I think we should be careful about this, because now quay uses `repository` as the permission area, but with `registry` added, it should maybe either use both interchangeably or switch to `registry`.

pkg/serviceprovider/gitlab/gitlab.go
```diff
@@ -0,0 +1,279 @@
...
+
+	return ret, nil
+}
+
+type gitlabProbe struct{}
+
+var _ serviceprovider.Probe = (*gitlabProbe)(nil)
+
+func (p gitlabProbe) Examine(_ *http.Client, url string) (string, error) {
+	// TODO: improve logic
```
metlos, line 274: Do you have some ideas how to probe the host to figure out if it is a gitlab server?

pkg/serviceprovider/gitlab/gitlabclientbuilder.go
```diff
@@ -0,0 +1,30 @@
...
+}
+
+func (g *gitlabClientBuilder) createAuthenticatedGlClient(ctx context.Context, spiAccessToken *api.SPIAccessToken, baseUrl string) (*gitlab.Client, error) {
+	lg := log.FromContext(ctx)
+	tokenData, err := g.tokenStorage.Get(ctx, spiAccessToken)
+	if err != nil {
+		lg.Error(err, "failed to get token from storage for", "token", spiAccessToken)
+		return nil, fmt.Errorf("failed to get token from storage for %s/%s: %w", spiAccessToken.Namespace, spiAccessToken.Name, err)
+	}
+	lg.V(logs.DebugLevel).Info("Created new github client", "SPIAccessToken", spiAccessToken)
```
metlos, line 27: `Create a new gitlab client` :)

pkg/serviceprovider/gitlab/metadataprovider.go
```diff
@@ -0,0 +1,108 @@
...
+		lg.Error(err, "error during fetching user metadata from GitLab", "status", resp.StatusCode)
+		return "", "", nil, nil
+	}
+
+	type tokenResponseBody struct {
+		scope []string
+	}
+	responseBody := tokenResponseBody{}
+
+	req, err := retryablehttp.NewRequestWithContext(ctx, "GET", p.baseUrl+"/oauth/token/info", nil)
```
metlos, line 88: Hmm... so the scopes of the user are "global" and don't differ per-repository?

api/v1beta1/spiaccesstoken_types.go
xbaran4: I agree.

pkg/serviceprovider/gitlab/gitlab.go
xbaran4: 1. just regex match - something like 
- gitlab.service.com = yes
- service.domain.com/gitlab-repo = false
2. make some request to the URL that should respond in gitlab specific way
3. try to match the host part to some gitlab record in user config
4. others...

pkg/serviceprovider/gitlab/metadataprovider.go
xbaran4: If you mean something like the Viewer permission on GitHub, I haven't seen something like it on GitLab.

pkg/serviceprovider/gitlab/gitlab.go
```diff
@@ -0,0 +1,308 @@
...
+	if len(tokens) == 0 {
+		return nil, nil
+	}
+
+	return &tokens[0], nil
+}
+
+func (g Gitlab) PersistMetadata(ctx context.Context, _ client.Client, token *api.SPIAccessToken) error {
+	if err := g.lookup.PersistMetadata(ctx, token); err != nil {
+		return fmt.Errorf("failed to persiste gitlab metadata: %w", err)
```
mshaposhnik, line 109: typo

pkg/serviceprovider/gitlab/gitlabclientbuilder.go
```diff
@@ -0,0 +1,44 @@
...
+
+func (builder *gitlabClientBuilder) createGitlabAuthClient(ctx context.Context, spiAccessToken *api.SPIAccessToken, baseUrl string) (*gitlab.Client, error) {
+	lg := log.FromContext(ctx)
+	tokenData, err := builder.tokenStorage.Get(ctx, spiAccessToken)
+	if err != nil {
+		lg.Error(err, "failed to get token from storage for", "token", spiAccessToken)
+		return nil, fmt.Errorf("failed to get token from storage for %s/%s: %w", spiAccessToken.Namespace, spiAccessToken.Name, err)
+	}
+	lg.V(logs.DebugLevel).Info("Created new gitlab client", "SPIAccessToken", spiAccessToken)
+	return gitlab.NewOAuthClient(tokenData.AccessToken, gitlab.WithHTTPClient(builder.httpClient), gitlab.WithBaseURL(baseUrl))
```
mshaposhnik, line 42: that works only with OAuth tokens, right ? Unfortunately, PAT seems not acceptable there and leads to 
```
:"Error","error":"failed to persiste gitlab metadata: metadata cache error: fetching token data: GET https://gitlab.com/oauth/token/info: 401 {error: invalid_token}, {error_description: The access token is invalid}
```

mshaposhnik:
Works for me in general. 
image

Some notes:
- PAT-s need to be supported asap...
- `read_api` permission seems to be obligatory for the application (screenshot shows it off, but i got the error w/out it)
- `ServiceProviderState` value seems to be `{}` in base64

pkg/serviceprovider/gitlab/metadataprovider.go
metlos: So how are we going to figure out if a user has access to one repository as opposed to another?

pkg/serviceprovider/gitlab/gitlab.go
metlos: Yep, something like that... Have you already tried to find some kind of a gitlab-specific request like that?

api/v1beta1/spiaccesstoken_types.go
xbaran4: Issue created: https://issues.redhat.com/browse/SVPI-248

pkg/serviceprovider/gitlab/gitlab.go
xbaran4: The agreed-upon solution is as follows: The probe should have access to the service provider configuration and the configuration from custom secrets as in https://github.com/redhat-appstudio/service-provider-integration-operator/pull/302. Then compare the base of repoUrl from biding to the baseUrls off all GitLab service provider configurations.

pkg/serviceprovider/gitlab/gitlab.go
xbaran4: Fixed.

pkg/serviceprovider/gitlab/metadataprovider.go
xbaran4: You are right. I will have to do more research on how this can be achieved with GitLab and implement it into TokenState. However, as this is supposed to be an MVP, I would like to merge this PR without it and afterward create a new PR with the implementation.

sparkoo (commented):
code looks fine, just few minor comments. tbh I haven't tested it in the real deployment.

However, I think you've really overscoped this. whole oauth part and spiaccesscheck part was not necessary to do in this single PR. It makes it harder to test and review and makes it more prone to bugs.

sparkoo (approved):

PR #320, opened 2022-10-27
Docs porting

mshaposhnik:
### What does this PR do?
Port some docs from google drive to git

https://issues.redhat.com/browse/SVPI-220

### How to test this PR?

Just read the docs & enjoy : -)


sparkoo (approved):
I think it's good first step to move the docs here and I'm ok to merge it as is. However, I have few major comments, but all can be done as followups:
 - we've talked before about writing the docs for various actors - user guide, admin guide, ... Have you been thinking about that?
 - If we want to have CRD docs, I think it should be generated from yamls

Can you please also make sure, after we merge this in any form, to write some visible comment to source documents, that it was moved to gh repo + link?

metlos (approved):

PR #336, opened 2022-11-10
Remove the limitation for one service provider per type in the oauth service

metlos:
### What does this PR do?
The URLs for initiating the OAuth flow needlessly contained the service provider type which made it impossible to have more than 1 service provider per type.

This PR removes that limitation

none

### How to test this PR?
TBD


sparkoo:
I'm closing this in favor of https://github.com/redhat-appstudio/service-provider-integration-operator/pull/341

PR #337, opened 2022-11-14
Allow file content requests from GitLab servers

mshaposhnik:
### What does this PR do?

Allows to create & deliver content for FileContentRequests for GitLab repositories.

https://issues.redhat.com/browse/SVPI-253
https://issues.redhat.com/browse/SVPI-242

### How to test this PR?

 Deploy & create File request CR as follows:
 
 ```
kind: SPIFileContentRequest
metadata:
  name: test-file-content-request
  namespace: default
spec:
  repoUrl: https://gitlab.com/mshaposhnik/proj1
  filePath: foo/bar.txt
```

provide token either via OAuth or upload. Verify file content is injected.

### CODE REVIEW

**Comment by mshaposhnik** [comment]
> It might be also a good idea to think about moving this logic out from the controller. Or at least part of it.
> 
> ```
> 			url, err := downloadableSp.GetFileDownloadUrl(ctx, request.Spec.RepoUrl, request.Spec.FilePath, request.Spec.Ref, token)
> 			if err != nil {
> 				lg.Error(err, "fail to calculate the file download URL")
> 				r.updateFileRequestStatusError(ctx, &request, err)
> 				return ctrl.Result{}, fmt.Errorf("unable to calculate the file download URL: %w", err)
> 			}
> 			contents, err := gitfile.GetFileContents(ctx, r.K8sClient, *r.HttpClient, url, request.Namespace, binding.Status.SyncedObjectRef.Name)
> 			if err != nil {
> 				r.updateFileRequestStatusError(ctx, &request, err)
> 				return reconcile.Result{}, fmt.Errorf("error fetching file content: %w", err)
> 			}
> 			fileBytes, err := io.ReadAll(contents)
> 			if err != nil {
> 				r.updateFileRequestStatusError(ctx, &request, err)
> 				return reconcile.Result{}, fmt.Errorf("error reading file content: %w", err)
> 			}
> ```

File: pkg/serviceprovider/serviceprovider.go [hunk 1]
```diff
@@ -170,6 +170,13 @@ type Matchable interface {
 	ObjNamespace() string
 }
 
+type Downloadable interface {
```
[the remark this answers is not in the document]
mshaposhnik: Not sure if i get idea fully.... gitHub and Gitlab will implement the `ScmProvider`, others are not, right ?
how does particular fecture support check will looks like ?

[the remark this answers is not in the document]
mshaposhnik: what you saying is that general check, but not for particular feature. We will not be adble to disctinct between  particular option support if there will be more of them.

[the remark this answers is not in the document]
mshaposhnik: Ok rewrite in similar way

pkg/serviceprovider/gitlab/downloadfilecapability.go
```diff
@@ -0,0 +1,93 @@
...
+	unexpectedStatusCodeError  = errors.New("unexpected status code from GitLab API")
+	fileSizeLimitExceededError = errors.New("failed to retrieve file: size too big")
+)
+
+var maxFileSizeLimit int = 2097152
+
+func (f downloadFileCapability) DownloadFile(ctx context.Context, repoUrl, filepath, ref string, token *api.SPIAccessToken) (string, error) {
+	gitLabURLRegexpNames := f.gitLabUrlRegexp.SubexpNames()
+	result := f.gitLabUrlRegexp.FindAllStringSubmatch(repoUrl, -1)
+	m := map[string]string{}
```
sparkoo, line 59: same as above, I would very welcome better variable names

controllers/spifilecontentrequest_controller.go
```diff
@@ -182,18 +182,34 @@ func (r *SPIFileContentRequestReconciler) Reconcile(ctx context.Context, req ctr
...
 			request.Status.TokenUploadUrl = binding.Status.UploadUrl
 		} else if binding.Status.Phase == api.SPIAccessTokenBindingPhaseInjected {
-			contents, err := gitfile.GetFileContents(ctx, r.K8sClient, *r.HttpClient, request.Namespace, binding.Status.SyncedObjectRef.Name, request.Spec.RepoUrl, request.Spec.FilePath, request.Spec.Ref)
+			sp, err := r.ServiceProviderFactory.FromRepoUrl(ctx, request.Spec.RepoUrl)
 			if err != nil {
-				r.updateFileRequestStatusError(ctx, &request, err)
-				return reconcile.Result{}, fmt.Errorf("error fetching file content: %w", err)
+				lg.Error(err, "unable to get the service provider")
+				// we determine the service provider from the URL in the spec. If we can't do that, nothing works until the
+				// user fixes that URL. So no need to repeat the reconciliation and therefore no error returned here.
```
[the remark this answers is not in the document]
mshaposhnik: Added error status

pkg/serviceprovider/github/downloadfilecapability.go
```diff
@@ -0,0 +1,77 @@
...
+)
+
+var _URLRegexp = regexp.MustCompile(`(?Um)^(?:https)(?:\:\/\/)github.com/(?P<owner>[^/]+)/(?P<repo>[^/]+)(.git)?$`)
+var _URLRegexpNames = _URLRegexp.SubexpNames()
+var maxFileSizeLimit int = 2097152
+
+func (f downloadFileCapability) DownloadFile(ctx context.Context, repoUrl, filepath, ref string, token *api.SPIAccessToken) (string, error) {
+	result := _URLRegexp.FindAllStringSubmatch(repoUrl, -1)
+	m := map[string]string{}
+	for i, n := range result[0] {
```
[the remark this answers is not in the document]
mshaposhnik: i'll try

controllers/spifilecontentrequest_controller.go
```diff
@@ -154,13 +154,24 @@ func (r *SPIFileContentRequestReconciler) Reconcile(ctx context.Context, req ctr
...
+			return ctrl.Result{}, serviceprovider.FileDownloadNotSupportedError{}
+		}
 		request.Status.Phase = api.SPIFileContentRequestPhaseAwaitingTokenData
 	} else if request.Status.Phase == api.SPIFileContentRequestPhaseDelivered {
 		// already injected, nothing to do
 		return ctrl.Result{}, nil
 	}
 
 	if request.Status.LinkedBindingName == "" {
+		if request.Status.Phase == api.SPIFileContentRequestPhaseError {
```
metlos, line 174: Is it possible that we can recover from the state where we set the phase to error? if so, we should not short-circuit here...

controllers/spifilecontentrequest_controller.go
```diff
@@ -182,18 +193,35 @@ func (r *SPIFileContentRequestReconciler) Reconcile(ctx context.Context, req ctr
...
 		} else if binding.Status.Phase == api.SPIAccessTokenBindingPhaseInjected {
-			contents, err := gitfile.GetFileContents(ctx, r.K8sClient, *r.HttpClient, request.Namespace, binding.Status.SyncedObjectRef.Name, request.Spec.RepoUrl, request.Spec.FilePath, request.Spec.Ref)
+			sp, err := r.ServiceProviderFactory.FromRepoUrl(ctx, request.Spec.RepoUrl)
 			if err != nil {
-				r.updateFileRequestStatusError(ctx, &request, err)
-				return reconcile.Result{}, fmt.Errorf("error fetching file content: %w", err)
+				lg.Error(err, "unable to get the service provider")
+				// we determine the service provider from the URL in the spec. If we can't do that, nothing works until the
+				// user fixes that URL. So no need to repeat the reconciliation and therefore no error returned here.
+				r.updateFileRequestStatusError(ctx, &request, fmt.Errorf("unable to find matchig service provider for the fiven URL"))
```
metlos: `... to find a matching service provider for the URL`

controllers/spifilecontentrequest_controller.go
```diff
@@ -182,18 +193,35 @@ func (r *SPIFileContentRequestReconciler) Reconcile(ctx context.Context, req ctr
...
+			downloadableSp, ok := sp.(serviceprovider.ScmProvider)
+			if !ok {
+				r.updateFileRequestStatusError(ctx, &request, serviceprovider.FileDownloadNotSupportedError{})
+				return ctrl.Result{}, serviceprovider.FileDownloadNotSupportedError{}
+			}
+
+			token := &api.SPIAccessToken{}
+			err = r.K8sClient.Get(ctx, client.ObjectKey{Namespace: request.Namespace, Name: binding.Status.LinkedAccessTokenName}, token)
+			if err != nil {
+				lg.Error(err, "unable to fetch the token")
```
metlos, line 216: Update the status here, too?

pkg/serviceprovider/scmprovider.go
```diff
@@ -0,0 +1,22 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package serviceprovider
+
+//ScmProvider is a marker interface for service providers, indicating that given SP is able to
+// perform all listed SCM-specific operations, such as file retrieving or others.
+type ScmProvider interface {
+	// GetDownloadFileCapability returns download file capability
+	GetDownloadFileCapability() DownloadFileCapability
```
metlos, line 21: What if the `ServiceProvider` interface had this method and returned a pointer to `DownloadCapability`? Returning `nil` would mean the service provider doesn't support file download.

That way we would avoid the checks by type-casting and maybe also simplify the situation in the testsuite where you would be simply able to make the `TestServiceProvider` support or not support file download...

pkg/serviceprovider/github/downloadfilecapability.go
```diff
@@ -0,0 +1,77 @@
...
+
+var (
+	unexpectedStatusCodeError  = errors.New("unexpected status code from GitHub API")
+	fileSizeLimitExceededError = errors.New("failed to retrieve file: size too big")
+	pathIsADirectoryError      = errors.New("provided path refers to a directory, not file")
+)
+
+var _URLRegexp = regexp.MustCompile(`(?Um)^(?:https)(?:\:\/\/)github.com/(?P<owner>[^/]+)/(?P<repo>[^/]+)(.git)?$`)
+var _URLRegexpNames = _URLRegexp.SubexpNames()
+var maxFileSizeLimit int = 2097152
```
metlos, line 46: How did we choose 2MB? Could this be made configurable?

pkg/serviceprovider/gitlab/file_url_resolver.go
```diff
@@ -0,0 +1,75 @@
...
+}
+
+var _ serviceprovider.FileUrlResolver = (*fileUrlResolver)(nil)
+
+var (
+	unexpectedStatusCodeError  = errors.New("unexpected status code from GitLab API")
+	fileSizeLimitExceededError = errors.New("failed to retrieve file: size too big")
+)
+
+var maxFileSizeLimit int = 2097152
```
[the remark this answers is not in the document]
mshaposhnik: \+

pkg/serviceprovider/scmprovider.go
mshaposhnik: To avoid unnecesary stubbing for the providers that didn't have SCM capabilities

pkg/serviceprovider/github/downloadfilecapability.go
mshaposhnik: \+

pkg/serviceprovider/scmprovider.go
metlos: In that case, why the double indirection? What if ScmProvider directly had the `DownloadFile` method?

metlos: Although to be fair, I'm still more a fan of just extending the `ServiceProvider` interface (and keeping the `DownloadCapability` so that we can easily distinguish the SPs that support file download from the ones that do not).

sparkoo (approved):

pkg/config/config.go
```diff
@@ -44,6 +44,7 @@ type OperatorCliArgs struct {
 	TokenMatchPolicy            TokenPolicy   `arg:"--token-match-policy, env" default:"any" help:"The policy to match the token against the binding. Options:  'any', 'exact'."`
 	ApiExportName               string        `arg:"--kcp-api-export-name, env" default:"spi" help:"SPI ApiExport name used in KCP environment to configure controller with virtual workspace."`
 	DeletionGracePeriod         time.Duration `arg:"--deletion-grace-period, env" default:"2s" help:"The grace period between a condition for deleting a binding or token is satisfied and the token or binding actually being deleted."`
+	MaxFileDownloadSize         int           `arg:"--max-download-size, env" default:"2097152" help:"A maximum file size for file downloading from SCM capabilities supporting providers"`
```
sparkoo, line 47: can you please define the unit of the value in help comment, or even name variable like `--max-download-size-bytes` or sth like that

mshaposhnik: ok

PR #338, opened 2022-11-15
feat: improved gitlab probing

xbaran4:

### What does this PR do?
The GitLab Probe and its Examine method needs to know about every service provider configuration and based on that decide whether repoUrl is from GitLab.

In order minimalize the calls to k8s API and not complicate the Examine method, this listing is done prior to calling the Examine method which is done for every service provider.

Solves https://issues.redhat.com/browse/SVPI-256

### How to test this PR?

1. Deploy SPI with operator image `quay.io/pbaran/spio:SVPI-256-probing`
2. Create user config secret such as `/samples/sp-secret.yaml` with a labes `spi.appstudio.redhat.com/service-provider-type: GitLab`  and `spi.appstudio.redhat.com/service-provider-host: some.gitlab.url.com`. The clientId and clientSecret should not matter. 
3. Create 3 bindings (with different names) such as `samples/binding-write-repo-gitlab.yaml` with different repo URLs: 
- https://some.gitlab.url.com/path/to/repo
- https://gitlab.com/path/to/repo
- https://gitlab.url.com/path/to/repo
4. The first two bindings should be linked to GitLab SpiAccessToken with their respective service provider URLs. The last should be linked to SpiAccessToken with HostCredentials service provider, as it does not match any known GitLab configuration.

5. IMPORTANT ADDITION. this PR now changes the shared code for service providers. Therefore you should also test creating GitHub and Quay bindings. You can use `samples/binding-write-repo.yaml` for GH and `samples/binding-quay-without-protocol.yaml` for Quay.

If there is a request to somehow limit this PR only for GitLab and introduce this change for every SP later, I am able to grant this.

### DIFF
--- a/controllers/spiaccesscheck_controller.go
+++ b/controllers/spiaccesscheck_controller.go
@@ -75,7 +75,7 @@ func (r *SPIAccessCheckReconciler) Reconcile(ctx context.Context, req ctrl.Reque
 		}
 	}
 
-	if sp, spErr := r.ServiceProviderFactory.FromRepoUrl(ctx, ac.Spec.RepoUrl); spErr == nil {
+	if sp, spErr := r.ServiceProviderFactory.FromRepoUrl(ctx, ac.Spec.RepoUrl, req.Namespace); spErr == nil {
 		auditLog := log.FromContext(ctx, "audit", "true", "namespace", ac.Namespace, "token", ac.Name, "repository", ac.Spec.RepoUrl)
 		auditLog.Info("performing repository access check")
 		if status, repoCheckErr := sp.CheckRepositoryAccess(ctx, r.Client, &ac); repoCheckErr == nil {
--- a/controllers/spiaccesstoken_controller.go
+++ b/controllers/spiaccesstoken_controller.go
@@ -193,7 +193,7 @@ func (r *SPIAccessTokenReconciler) Reconcile(ctx context.Context, req ctrl.Reque
 	}
 
 	// persist the SP-specific state so that it is available as soon as the token flips to the ready state.
-	sp, err := r.ServiceProviderFactory.FromRepoUrl(ctx, at.Spec.ServiceProviderUrl)
+	sp, err := r.ServiceProviderFactory.FromRepoUrl(ctx, at.Spec.ServiceProviderUrl, req.Namespace)
 	if err != nil {
 		if uerr := r.flipToExceptionalPhase(ctx, &at, api.SPIAccessTokenPhaseError, api.SPIAccessTokenErrorReasonUnknownServiceProvider, err); uerr != nil {
 			return ctrl.Result{}, fmt.Errorf("failed update the status: %w", uerr)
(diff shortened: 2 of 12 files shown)


metlos (commented):
I think that while this implementation does what it says, it is not completely correct from "architectural" point of view. 

By adding the map of known base URLs to the Examine methods of probes, you changed their purpose from trying to figure out if a certain URL is backed by certain SP to a lookup in the provided map. 

Or rather, different probes behave differently now - github and quay check the provided URL and disregard the provided map, meaning that they will return a non-empty result even if there is no github or quay configured, while gitlab will only consider the entries in the map.

I think we should keep the purpose of the probes to "probe" given URLs to see what kind of  SP might be handling them and lift the lookup in the map "above" the probes, i.e. IMHO it could be part of the `FromRepoUrl` logic (where you correctly added the namespace param so that this method can do the full lookup of available configured SPs).

This could technically mean that the probes will never get called, which I think is ok :) We will still need the probes at the point when we will either confirm the SP type given some configuration, or, given the recent discussions, when we provide "configurationless" SPs - just give us a URL and we figure out the rest.

xbaran4:
> Or rather, different probes behave differently now - github and quay check the provided URL and disregard the provided map, meaning that they will return a non-empty result even if there is no github or quay configured, while gitlab will only consider the entries in the map.

Because I am adding `https://gitlab.com` to that map the GitLab doesn't return empty result in case no GitLab is configured. But I understand your point.

> I think we should keep the purpose of the probes to "probe" given URLs to see what kind of SP might be handling them and lift the lookup in the map "above" the probes, i.e. IMHO it could be part of the `FromRepoUrl` logic (where you correctly added the namespace param so that this method can do the full lookup of available configured SPs).

Do you think it makes sense to generalize this PR, so that, initializers of every type look at the configured SPs?

> We will still need the probes at the point when we will either confirm the SP type given some configuration, or, given the recent discussions, when we provide "configurationless" SPs - just give us a URL and we figure out the rest.

If we keep the logic in `FromRepoUrl` the GitLab Probe for know can be empty right? Until we decide about what you mentioned above.

xbaran4:
> > > I think we should keep the purpose of the probes to "probe" given URLs to see what kind of SP might be handling them and lift the lookup in the map "above" the probes, i.e. IMHO it could be part of the `FromRepoUrl` logic (where you correctly added the namespace param so that this method can do the full lookup of available configured SPs).
> > 
> > 
> > Do you think it makes sense to generalize this PR, so that, initializers of every type look at the configured SPs?
> 
> No I think this logic should be in the factory - it would find the correct base URL of the service provider given the list of configured SPs in given namespace (ie. the ones from the config file + the ones from the namespace), do the URL match on them and pick the appropriate initializer which would just instantiate the SP instance.
> 
> > > We will still need the probes at the point when we will either confirm the SP type given some configuration, or, given the recent discussions, when we provide "configurationless" SPs - just give us a URL and we figure out the rest.
> > 
> > 
> > If we keep the logic in `FromRepoUrl` the GitLab Probe for now can be empty right? Until we decide about what you mentioned above.
> 
> Yes. We can introduce the advanced logic of "sniffing" the SP from a URL only when we get to that point.

Thank you so much for the useful feedback.

metlos (approved):
LGTM apart from the one comment below.

sparkoo (approved):

xbaran4:
Issue about public URLs https://issues.redhat.com/projects/SVPI/issues/SVPI-271?filter=allissues

PR #339, opened 2022-11-15
Simplify the reconcile methods of the controllers 

metlos:
### What does this PR do?
Simplify the reconcile methods of the controllers  to not rely so heavily on the status but rather derive all the state from the specs.

https://issues.redhat.com/browse/SVPI-216

### How to test this PR?
There should be no observable changes as opposed to the current function of the operator.


pkg/serviceprovider/serviceprovider.go
```diff
@@ -43,7 +43,7 @@ type ServiceProvider interface {
 	// LookupToken tries to match an SPIAccessToken object with the requirements expressed in the provided binding.
 	// This usually searches kubernetes (using the provided client) and the service provider itself (using some specific
 	// mechanism (usually an http client)).
-	LookupToken(ctx context.Context, cl client.Client, binding *api.SPIAccessTokenBinding) (*api.SPIAccessToken, error)
+	LookupToken(ctx context.Context, cl client.Client, binding *api.SPIAccessTokenBinding) ([]api.SPIAccessToken, error)
```
xbaran4, line 46: Could we change the name to `LookupTokens` and/or adjust the documentation above?

xbaran4 (approved):

PR #340, opened 2022-11-16
Extra gosec checks

skabashnyuk:
### What does this PR do?
This pr added extra redundancy to what we already have with golangci-lint-action such as:
 - Explicit linter check. This check is focused on gosec only and not mixed with any other thing.
 - On each update and by schedule upload sarif report  to the github https://docs.github.com/en/code-security/code-scanning/integrating-with-code-scanning/sarif-support-for-code-scanning

n/a

https://issues.redhat.com/browse/SVPI-265

### How to test this PR?
n/a

### DIFF
--- a/controllers/labels.go
+++ b/controllers/labels.go
@@ -14,4 +14,4 @@
 
 package controllers
 
-const SPIAccessTokenLinkLabel = "spi.appstudio.redhat.com/linked-access-token"
+const SPIAccessTokenLinkLabel = "spi.appstudio.redhat.com/linked-access-token" //#nosec G101 -- false positive, this is not a private data


.github/workflows/linter.yaml
```diff
@@ -31,3 +31,14 @@ jobs:
...
           # skip-build-cache: true
+  gosec:
+    runs-on: ubuntu-latest
+    env:
+      GO111MODULE: on
+    steps:
+      - name: Checkout Source
+        uses: actions/checkout@v3
+      - name: Run Gosec Security Scanner
+        uses: securego/gosec@master
```
sparkoo, line 42: is it good idea to run from master?

.github/workflows/sec.yaml
```diff
@@ -0,0 +1,27 @@
+name: "Security Scan"
+
+# Run workflow each time code is pushed to your repository and on a schedule.
+# The scheduled workflow runs every at 00:00 on Sunday UTC time.
+on:
+  push:
```
sparkoo, line 6: does this run on push or this action is run weekly?

.github/workflows/sec.yaml
```diff
@@ -0,0 +1,27 @@
...
+  schedule:
+    - cron: '0 0 * * 0'
```
sparkoo, line 8:
```suggestion
    - cron: '0 0 * * 0' # every Sunday 00:00 UTC
```

.github/workflows/sec.yaml
```diff
@@ -0,0 +1,27 @@
+name: "Security Scan"
+
+# Run workflow each time code is pushed to your repository and on a schedule.
+# The scheduled workflow runs every at 00:00 on Sunday UTC time.
```
sparkoo, line 4: I would just remove this whole comment and write inline comment to cron time

.github/workflows/linter.yaml
```diff
@@ -31,3 +31,14 @@ jobs:
 
           # Optional: if set to true then the action don't cache or restore ~/.cache/go-build.
           # skip-build-cache: true
+  gosec:
```
sparkoo, line 34: what level of security issue makes the whole check red?

metlos: I think these just might be caused the fact that you're using the latest greatest version of the github action, while our lint task pins the version of golang-lint. What about trying to change this https://github.com/redhat-appstudio/service-provider-integration-operator/blob/315765437588437b7cc9fb2a9ce1efd6c7764ecd/.github/workflows/linter.yaml#L18 to `latest` and see if we get the same new warnings?

metlos (commented):
I would personally rather modify our existing linting infra than to add another github action that basically duplicates what we already should have done in the linter.
If there is no other way, then I'm ok with this but I'd like to first try to not have to have duplicate work done.

sparkoo (approved):

PR #344, opened 2022-11-22
feat: changing quay repository area to registry area

xbaran4:

### What does this PR do?
Changes the valid permission areas for quay binding from `repository` and `repositoryMetadata` to `registry` and `registryMetadata` respectively.

Adds migration logic to the SPIAccessTokenBinding reconciler that changes the old areas into the new ones. 

image

### How to test this PR?

1. Deploy with operator: `quay.io/pbaran/spio:SVPI-248-quay`
2. Create binding: 
```yaml
kind: SPIAccessTokenBinding
metadata:
  name: test-binding-quay-without-protocol
  namespace: default
spec:
  permissions:
    required:
      - type: rw
        area: repository
      - type: rw
        area: repositoryMetadata
  repoUrl: quay.io/repository/yourorganization/image
  secret:
```
3. Look at the created binding after a second or two and it should have changed permission areas to `registry` and `registryMetadata` and there should be no error with the binding.
4. Optionally, you can go through the oauth flow.


pkg/serviceprovider/quay/quay.go
```diff
@@ -329,11 +320,13 @@ func (q *Quay) MapToken(ctx context.Context, binding *api.SPIAccessTokenBinding,
 func (q *Quay) Validate(ctx context.Context, validated serviceprovider.Validated) (serviceprovider.ValidationResult, error) {
 	ret := serviceprovider.ValidationResult{}
 
-	userPermissionAreaRequested := false
 	for _, p := range validated.Permissions().Required {
-		if p.Area == api.PermissionAreaUser && !userPermissionAreaRequested {
-			ret.ScopeValidation = append(ret.ScopeValidation, userRelatedPermissionsNotSupportedError)
-			userPermissionAreaRequested = true
+		switch p.Area {
```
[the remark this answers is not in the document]
xbaran4: The idea is that only `PermissionAreaRegistry` and `PermissionAreaRegistryMetadata` are allowed. Every other permission will result in a validation error. The difference is that if the user, for example, specifies in binding 3x user area instead of one (which happens rarely, IMHO), he will get 3 validation errors saying that the user area is not supported. But the code is simpler, and I don't think it impacts the user experience.

controllers/spiaccesstokenbinding_controller.go
```diff
@@ -221,6 +221,28 @@ func (r *SPIAccessTokenBindingReconciler) Reconcile(ctx context.Context, req ctr
 		return ctrl.Result{}, nil
 	}
 
+	// Migration logic for quay bindings that use repository or repository area.
+	permissionChange := false
```
sparkoo, line 225: can we move this to a function?

controllers/spiaccesstokenbinding_controller.go
```diff
@@ -221,6 +221,28 @@ func (r *SPIAccessTokenBindingReconciler) Reconcile(ctx context.Context, req ctr
...
+			}
+			if permission.Area == api.PermissionAreaRepositoryMetadata {
+				binding.Spec.Permissions.Required[i].Area = api.PermissionAreaRegistryMetadata
+				permissionChange = true
+			}
+		}
+	}
+	if permissionChange {
+		lg.Info("migrating old permission areas for quay", "binding", binding)
+		if err = r.Client.Update(ctx, &binding); err != nil {
```
sparkoo, line 240: shouldn't this return and retrigger reconcile ?

controllers/spiaccesstokenbinding_controller.go
```diff
@@ -221,6 +221,28 @@ func (r *SPIAccessTokenBindingReconciler) Reconcile(ctx context.Context, req ctr
 		return ctrl.Result{}, nil
 	}
 
+	// Migration logic for quay bindings that use repository or repository area.
```
sparkoo, line 224: shouldn't it be in the beginning of reconcile ?

metlos (commented):
I have nothing to add on top of @sparkoo's review.

PR #360, opened 2022-12-02
add tokenUpload controller

gazarenkov:
### What does this PR do?
It enables upload of Personal Access Token using very short living K8s Secret.
Controller recognizes the Secret by label.spi.appstudio.redhat.com/token: "" label, gets the PAT and the Name of the SPIAccessToken associated with it, then deletes the Secret (for better security reason) and uploads the Token (which in turn updates the Status of associated SPIAccess/Token/TokenBinding).

https://issues.redhat.com/browse/SVPI-273

### How to test this PR?
No automated tests provided for the time (TODO)
To test it manually:
- Run operator with  ENABLETOKENUPLOAD=true. For example off-cluster it may look like:  
make run_as_current_user VAULTHOST=https://vault.192.168.64.6.nip.io VAULTAPPROLEROLEIDFILEPATH=.secretdata/role_id VAULTAPPROLESECRETIDFILEPATH=.secretdata/secret_id VAULTINSECURETLS=true CONFIGFILE=.secretdata/config.yaml ENABLETOKENUPLOAD=true
- Create spiaccesstokenbinding object with SP of your choice. For example: spec.repoUrl: https://github.com for GitHub 
- Check this spiaccesstokenbinding, like: kubectl get spiaccesstokenbinding <name> -o yaml . It's status.phase should be AwaitingTokenData
- Take Linked Token Name from status.linkedAccessTokenName. 
- Get Service Provider's Personal Access Token. For example from GitHub->Settings->Developer Settings->Personal Access Tokens
- Create the Secret like:
```
apiVersion: v1
kind: Secret
metadata:
  name: test-secret
  labels:
    spi.appstudio.redhat.com/upload-secret: token
    spi.appstudio.redhat.com/token-name: <spi-access-token-name>
type: Opaque
stringData:
  tokenData: <serviceprovider-pat>
```
- Check this spiaccesstokenbinding, like: kubectl get spiaccesstokenbinding <name> -o yaml . It's status.Phase should be Injected


gazarenkov:
> I wasn't able to upload the token with such a secret
> 
> ```
> cat <<EOF | kubectl apply -n $TARGET_NAMESPACE -f -
> apiVersion: v1
> kind: Secret
> metadata:
>   name: github-token
>   labels:
>     label.spi.appstudio.redhat.com/token: ""
> type: Opaque    
> stringData:                                                                                                                                                          
>   spiTokenName: ${LINKED_TOKEN_NAME}
>   tokenData: ${GITHUB_SPI}
> EOF
> ```
> 
> please correct me if I'm wrong.

The label should be: **spi.appstudio.redhat.com/token: ""** 
(w/o label. )otherwise it looks good

> 
> Also, could we reuse this function
> 
> https://github.com/redhat-appstudio/service-provider-integration-operator/blob/130d81f328cf6e3ab794ad625095a92a1b0161fb/oauth/token_upload.go#L51-L61
> 
> 
> or at least we should not forget to add audit logs.

Sure, let's discuss where we need to use it (I am not sure)

controllers/tokenupload_controller.go
```diff
@@ -0,0 +1,229 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/log"
+
+	spi "github.com/redhat-appstudio/service-provider-integration-operator/api/v1beta1"
+
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+const tokenSeretLabel = "spi.appstudio.redhat.com/token"
```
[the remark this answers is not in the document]
gazarenkov: Well, it can be "spi.appstudio.redhat.com/secret-type"="token"  , I'd say it is pretty much the same.
On one hand, I agree it looks nicer to have not empty label value, on the other hand I am not sure we use a "secret" word in our project vocabulary, unlike "token" word. So It may be some confusing from terminology standpoint...
WDYT?

As for moving spiTokenName (and providerURL too) to the label, we can do that since it probably fits the labels intention: "to be used to specify identifying attributes of objects that are meaningful and relevant to users, but do not directly imply semantics to the core system." 
+1

controllers/tokenupload_controller.go
```diff
@@ -0,0 +1,229 @@
...
+
+// TokenUploadReconciler reconciles a Secret object
+type TokenUploadReconciler struct {
+	client.Client
+	Scheme       *runtime.Scheme
+	TokenStorage tokenstorage.NotifyingTokenStorage
+}
+
+func (r *TokenUploadReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	//_ = log.FromContext(ctx)
```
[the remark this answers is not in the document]
gazarenkov: +1

controllers/tokenupload_controller.go
```diff
@@ -0,0 +1,229 @@
...
+		lg.Error(err, "Can not get list of secrets ")
+		return ctrl.Result{}, err
+	}
+
+	for _, s := range list.Items {
+
+		var accessToken spi.SPIAccessToken
+
+		// we immediatelly delete the Secret
+		err := r.Delete(context.TODO(), &s)
```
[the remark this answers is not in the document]
gazarenkov: Agreed, thanks

pkg/config/config.go
```diff
@@ -45,6 +45,8 @@ type OperatorCliArgs struct {
 	ApiExportName               string        `arg:"--kcp-api-export-name, env" default:"spi" help:"SPI ApiExport name used in KCP environment to configure controller with virtual workspace."`
 	DeletionGracePeriod         time.Duration `arg:"--deletion-grace-period, env" default:"2s" help:"The grace period between a condition for deleting a binding or token is satisfied and the token or binding actually being deleted."`
 	MaxFileDownloadSize         int           `arg:"--max-download-size-bytes, env" default:"2097152" help:"A maximum file size in bytes for file downloading from SCM capabilities supporting providers"`
+
```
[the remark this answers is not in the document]
gazarenkov: +1

controllers/tokenupload_controller.go
```diff
@@ -0,0 +1,229 @@
...
+}
+
+func (r *TokenUploadReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	//_ = log.FromContext(ctx)
+
+	lg := log.FromContext(ctx)
+
+	list := &corev1.SecretList{}
+
+	err := r.List(context.TODO(), list, client.HasLabels{tokenSeretLabel})
```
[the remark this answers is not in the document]
gazarenkov: will be fixed, thx

controllers/tokenupload_controller.go
```diff
@@ -0,0 +1,229 @@
...
+		err := r.Delete(context.TODO(), &s)
+		if err != nil {
+			logError(s, err, r, "can not delete the Secret ", lg)
+			return ctrl.Result{}, err
+		}
+
+		// if spiTokenName field is not empty - try to find SPIAccessToken by it
+		if len(s.Data["spiTokenName"]) > 0 {
+			accessToken = spi.SPIAccessToken{}
+			err = r.Get(context.TODO(), types.NamespacedName{Name: string(s.Data["spiTokenName"]), Namespace: s.Namespace}, &accessToken)
```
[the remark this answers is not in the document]
gazarenkov: will be fixed, thx

controllers/tokenupload_controller.go
```diff
@@ -0,0 +1,229 @@
...
+
+			if err != nil {
+				logError(s, err, r, "can not find SPI access token "+string(s.Data["spiTokenName"]), lg)
+				return ctrl.Result{}, err
+			} else {
+				lg.Info("SPI Access Token found : " + accessToken.Name)
```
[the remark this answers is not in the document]
gazarenkov: Do you mean Info(...) ?

controllers/tokenupload_controller.go
```diff
@@ -0,0 +1,229 @@
...
+		lg.Info("event Found and will be deleted: " + stored.Name)
+		err = r.Delete(context.TODO(), stored)
+		if err != nil {
+			lg.Error(err, " can not delete Event ")
+		}
+
+	}
+}
+
+func findTokenByUrl(url string, r *TokenUploadReconciler, lg logr.Logger) *spi.SPIAccessToken {
```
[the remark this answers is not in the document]
gazarenkov: +1

controllers/tokenupload_controller.go
```diff
@@ -0,0 +1,229 @@
...
+
+		if err1 != nil {
+			lg.Error(err1, "Event creation failed ")
+		}
+	}
+
+}
+
+func tryDeleteEvent(secretName string, ns string, r *TokenUploadReconciler, lg logr.Logger) {
+	stored := &corev1.Event{}
```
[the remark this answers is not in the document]
gazarenkov: +1

controllers/tokenupload_controller.go
mshaposhnik, line 42:
```suggestion
const tokenSecretLabel = "spi.appstudio.redhat.com/token"
```

controllers/tokenupload_controller.go
```diff
@@ -0,0 +1,229 @@
...
+		}
+
+		// if spiTokenName field is not empty - try to find SPIAccessToken by it
+		if len(s.Data["spiTokenName"]) > 0 {
+			accessToken = spi.SPIAccessToken{}
+			err = r.Get(context.TODO(), types.NamespacedName{Name: string(s.Data["spiTokenName"]), Namespace: s.Namespace}, &accessToken)
+
+			if err != nil {
+				logError(s, err, r, "can not find SPI access token "+string(s.Data["spiTokenName"]), lg)
+				return ctrl.Result{}, err
```
mshaposhnik, line 82: Can we consider `continue` operator here? Otherwise we can be completely blocked by one "bad" secret.

controllers/tokenupload_controller.go
gazarenkov: +

controllers/tokenupload_controller.go
gazarenkov: Well, it will not be blocked because the Secret is deleted so will not appear in the next reconcile cycle.
But, I agree, interrupting the loop this way is not nice and not optimal, thanks.

controllers/tokenupload_controller.go
```diff
@@ -0,0 +1,235 @@
...
+				lg.V(logs.DebugLevel).Info("SPI Access Token found : " + accessToken.Name)
+			}
+
+			// spiTokenName field is empty
+			// check providerUrl field and if not empty - try to find the token for this provider instance
+		} else if s.Labels[providerUrlLabel] != "" {
+			providerUrl := s.Labels[providerUrlLabel]
+			// NOTE: it does not fit advanced policy of matching token!
+			// Do we need it as an SPI "API function" which take into account this policy?
+			tkn := findTokenByUrl(ctx, providerUrl, r, lg)
```
mshaposhnik, line 99: can we limit that search by passing namespace known from secret? That will be much less expensive search imo..

controllers/tokenupload_controller.go
```diff
@@ -0,0 +1,235 @@
...
+	secretErrEvent.Type = "Error"
+	secretErrEvent.LastTimestamp = metav1.NewTime(time.Now())
+
+	err1 := r.Create(ctx, secretErrEvent)
+
+	if err1 != nil {
+		lg.Error(err1, "Event creation failed for Secret: "+secret.Name)
+	}
+
+	log.Log.Error(err, "Secret upload failed:")
```
mshaposhnik, line 203: `lg.Error` ?

controllers/tokenupload_controller.go
gazarenkov: @mshaposhnik not sure I understand. Could you please elaborate?

mshaposhnik: If i got it right, we searching tokens across all namespaces, but imo we need search only where the secret was uploaded. Or i am mistaken?

controllers/tokenupload_controller.go
```diff
@@ -0,0 +1,235 @@
...
+
+// SetupWithManager sets up the controller with the Manager.
+func (r *TokenUploadReconciler) SetupWithManager(mgr ctrl.Manager) error {
+	return ctrl.NewControllerManagedBy(mgr).
+		For(&corev1.Secret{}).
+		WithEventFilter(createTokenPredicate()).
+		Complete(r)
+}
+
+func createTokenPredicate() predicate.Predicate {
```
mshaposhnik, line 161: pls review that name

.secretdata/config.yaml
```diff
@@ -0,0 +1,10 @@
+serviceProviders:
```
mshaposhnik, line 1: is that test data accidentally committed?

mshaposhnik (approved):
LGTM in general with minor nitpicks inlined

cmd/operator/operator.go
```diff
@@ -102,6 +102,24 @@ func main() {
 		os.Exit(1)
 	}
 
+	// Setup tokenUpload controller if configured
+	// Important: need NotifyingTokenStorage to reconsile related SPIAccessToken
```
sparkoo, line 106:
```suggestion
	// Important: need NotifyingTokenStorage to reconcile related SPIAccessToken
```

controllers/tokenupload_controller.go
```diff
@@ -0,0 +1,235 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/log"
+
+	spi "github.com/redhat-appstudio/service-provider-integration-operator/api/v1beta1"
+
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+const tokenSecretLabel = "spi.appstudio.redhat.com/upload-secret"
```
sparkoo, line 45: can we make it block of constants like
```
const (
  tokenSecretLabel = ...
  spiTokenNameLabel = ...
  ...
)
```

controllers/tokenupload_controller.go
```diff
@@ -0,0 +1,235 @@
...
+const spiTokenNameLabel = "spi.appstudio.redhat.com/token-name"
+const providerUrlLabel = "spi.appstudio.redhat.com/providerUrl"
```
sparkoo, line 47: we use dashes in labels

controllers/tokenupload_controller.go
```diff
@@ -0,0 +1,235 @@
...
+	client.Client
+	Scheme       *runtime.Scheme
+	TokenStorage tokenstorage.NotifyingTokenStorage
+}
+
+func (r *TokenUploadReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+
+	lg := log.FromContext(ctx)
+
+	list := &corev1.SecretList{}
```
sparkoo, line 60: better variable name pls

controllers/tokenupload_controller.go
```diff
@@ -0,0 +1,235 @@
...
+
+	// list secrets labeled with "spi.appstudio.redhat.com/upload-secret"
+	err := r.List(ctx, list, client.HasLabels{tokenSecretLabel})
+	if err != nil {
+		lg.Error(err, "can not get list of secrets")
+		return ctrl.Result{}, fmt.Errorf("can not get list of secrets: %w,", err)
+	}
+
+	for i, s := range list.Items {
```
sparkoo, line 69: better var name

controllers/tokenupload_controller.go
```diff
@@ -0,0 +1,235 @@
...
+
+		var accessToken spi.SPIAccessToken
+
+		// we immediatelly delete the Secret
+		err := r.Delete(ctx, &list.Items[i])
```
sparkoo, line 74: why use `list.Items[i]` here and not secret variable ?

controllers/tokenupload_controller.go
```diff
@@ -0,0 +1,235 @@
...
+	secretErrEvent := &corev1.Event{}
+	secretErrEvent.Name = secret.Name
+	secretErrEvent.Message = err.Error()
+	secretErrEvent.Namespace = secret.Namespace
+	secretErrEvent.Reason = "Can not upload access token"
+	secretErrEvent.InvolvedObject = corev1.ObjectReference{Namespace: secret.Namespace, Name: secret.Name, Kind: secret.Kind, APIVersion: secret.APIVersion}
+	secretErrEvent.Type = "Error"
+	secretErrEvent.LastTimestamp = metav1.NewTime(time.Now())
+
+	err1 := r.Create(ctx, secretErrEvent)
```
sparkoo, line 197: better var name

controllers/tokenupload_controller.go
```diff
@@ -0,0 +1,235 @@
...
+
+		// we immediatelly delete the Secret
+		err := r.Delete(ctx, &list.Items[i])
+		if err != nil {
+			logError(ctx, s, fmt.Errorf("can not delete the Secret: %w", err), r, lg)
+			continue
+		}
+
+		// if spiTokenName field is not empty - try to find SPIAccessToken by it
+		if s.Labels[spiTokenNameLabel] != "" {
```
sparkoo, line 81: would be nice to have logic split to functions from the beginning

controllers/tokenupload_controller.go
sparkoo, line 45: I'm not convinced by this label name. should it rather be `upload-token`?

controllers/tokenupload_controller.go
sparkoo: what format should this url be?

controllers/tokenupload_controller.go
```diff
@@ -0,0 +1,235 @@
...
+				lg.V(logs.DebugLevel).Info("SPI Access Token found : " + accessToken.Name)
+			}
+
+			// spiTokenName field is empty
+			// check providerUrl field and if not empty - try to find the token for this provider instance
+		} else if s.Labels[providerUrlLabel] != "" {
+			providerUrl := s.Labels[providerUrlLabel]
+			// NOTE: it does not fit advanced policy of matching token!
+			// Do we need it as an SPI "API function" which take into account this policy?
+			tkn := findTokenByUrl(ctx, providerUrl, s.Namespace, r, lg)
```
sparkoo, line 99: should we somehow validate the url so it's in format we need?

sparkoo (commented):
it needs some care

- remove `.secretdata`
- cleanup variable names, formating, etc.
- tests
- readme

controllers/tokenupload_controller.go
gazarenkov: we do not use it as URL in the function, it is just a string to compare, so we can add one more validation to have a nice message in a case URL is not valid but it hardly bring any value IMO because the fact that it is not found (which should be the case b/c I suppose all Token's URLs are valid.

controllers/tokenupload_controller.go
gazarenkov: please propose what exactly you have in mind

controllers/tokenupload_controller.go
gazarenkov: :) well agree in general but why err is fine and err1 is not?

controllers/tokenupload_controller.go
gazarenkov: because: 
G601: Implicit memory aliasing in for loop. (gosec)
See skabashnyuk's suggestion above

controllers/tokenupload_controller.go
gazarenkov: please suggest the name

controllers/tokenupload_controller.go
gazarenkov: we can but I consider it a matter of taste.
is there any issue with keeping it as it is?

cmd/operator/operator.go
gazarenkov: +1

controllers/tokenupload_controller.go
gazarenkov: please suggest more appropriate name

.secretdata/config.yaml
gazarenkov: sure, thanks a lot!

controllers/tokenupload_controller.go
gazarenkov: OK, I change the logic to not to have err1 variable at all for simplicity

controllers/tokenupload_controller.go
gazarenkov: OK, remade for simplicity

controllers/spiaccesstokendataupdate_controller.go
```diff
@@ -96,6 +96,8 @@ func (r *SPIAccessTokenDataUpdateReconciler) Reconcile(ctx context.Context, req
 		return ctrl.Result{}, fmt.Errorf("failed to delete the processed data token update: %w", err)
 	}
 
+	lg.V(logs.DebugLevel).Info(fmt.Sprintf("token data update deleted: %s", req.Name))
```
sparkoo, line 99: do we want to keep it here ?

controllers/starter.go
```diff
@@ -89,5 +89,22 @@ func SetupAllReconcilers(mgr controllerruntime.Manager, cfg *config.OperatorConf
 		return err
 	}
 
+	// Setup tokenUpload controller if configured
+	// Important: need NotifyingTokenStorage to reconcile related SPIAccessToken
+	notifyingStorage := tokenstorage.NotifyingTokenStorage{
```
sparkoo, line 94: can we move it behind `cfg.EnableTokenUpload`

controllers/tokenupload_controller.go
```diff
@@ -0,0 +1,252 @@
...
+
+// reports error in both Log and Event in current Namespace
+func logError(ctx context.Context, secret corev1.Secret, err error, r *TokenUploadReconciler, lg logr.Logger) {
+
+	lg.Error(err, "Secret upload failed:")
+
+	tryDeleteEvent(ctx, secret.Name, secret.Namespace, r, lg)
+
+	secretErrEvent := &corev1.Event{}
+	secretErrEvent.Name = secret.Name
```
sparkoo, line 189: maybe inline initialization would look nicer? like:
```
secretErrEvent := &corev1.Event{
  Name: secret.Name,
  ...
}
```

controllers/tokenupload_controller.go
```diff
@@ -0,0 +1,252 @@
...
+
+	tokenList := spi.SPIAccessTokenList{}
+	err := r.List(ctx, &tokenList, client.InNamespace(ns))
+	if err != nil {
+		lg.Error(err, "Can not get tokenList of tokens ")
+		return nil
+	}
+
+	for _, token := range tokenList.Items {
+
```
sparkoo, line 228:
```suggestion
```

controllers/tokenupload_controller.go
```diff
@@ -0,0 +1,252 @@
...
+			lg.Error(err, " can not delete Event ")
+		}
+	}
+}
+
+func findTokenByUrl(ctx context.Context, url string, ns string, r *TokenUploadReconciler, lg logr.Logger) *spi.SPIAccessToken {
+
+	tokenList := spi.SPIAccessTokenList{}
+	err := r.List(ctx, &tokenList, client.InNamespace(ns))
+	if err != nil {
```
sparkoo, line 222:
```suggestion
	if err := r.List(ctx, &tokenList, client.InNamespace(ns)); err != nil {
```

controllers/tokenupload_controller.go
```diff
@@ -0,0 +1,252 @@
...
+			GenerateName: "generated-spi-access-token-",
+			Namespace:    ns,
+		},
+		Spec: spi.SPIAccessTokenSpec{
+			ServiceProviderUrl: providerUrl,
+		},
+	}
+	err := r.Create(ctx, &accessToken)
+	return accessToken, err
+
```
sparkoo, line 251:
```suggestion
```

controllers/tokenupload_controller.go
```diff
@@ -0,0 +1,252 @@
...
+		ObjectMeta: metav1.ObjectMeta{
+			GenerateName: "generated-spi-access-token-",
+			Namespace:    ns,
+		},
+		Spec: spi.SPIAccessTokenSpec{
+			ServiceProviderUrl: providerUrl,
+		},
+	}
+	err := r.Create(ctx, &accessToken)
+	return accessToken, err
```
sparkoo, line 250: do we want to return `accessToken` in case of error ?

gazarenkov: I think it makes no difference b/c we analyze err first of all, no?

controllers/tokenupload_controller.go
gazarenkov: +

controllers/spiaccesstokendataupdate_controller.go
gazarenkov: I'd prefer to have this debug info, can remove if critical for you

controllers/starter.go
gazarenkov: sure

controllers/spiaccesstokendataupdate_controller.go
sparkoo: ok, I'm fine with that

controllers/tokenupload_controller.go
sparkoo: In theory yes, but there might be different callers in the future and this is something I would not expect. Usual practice in go is that returned object is nil in case of error. But nothing critical.

sparkoo (approved):

controllers/tokenupload_controller.go
gazarenkov: OK, thanks, b/c I would not like to change the logic it depends on

PR #362, opened 2022-12-06
add the "ready" target to the Makefile

metlos:
### What does this PR do?
The goal should make the sources ready to pass the CI. It makes sure that the code is formatted, has license, lints correctly, passes the tests and go.mod is up-to-date.

none

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -149,6 +149,10 @@ lint: ## Run the linter on the codebase
 
 check: check_fmt lint test ## Check that the code conforms to all requirements for commit. Formatting, licenses, vet, tests and linters
 
+ready: fmt fmt_license lint test	
+	go mod download
+	go mod tidy
+
 vet: ## Run go vet against code.
 	go vet ./...
 
--- a/Makefile
+++ b/Makefile
@@ -101,10 +101,12 @@ help: ## Display this help.
 
 ##@ Development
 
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -153,6 +153,10 @@ lint: ## Run the linter on the codebase
 
 check: check_fmt lint test ## Check that the code conforms to all requirements for commit. Formatting, licenses, vet, tests and linters
 
+ready: fmt fmt_license lint test
```
sparkoo, line 156: does it make sense to run it with other tasks? like `test` maybe ?

sparkoo, line 156: wouldnt it make more sense to add `fmt_license` and `lint` to the `test` target?

sparkoo: or remove `test` from this and run `ready` from `test`

metlos: My original vision was to use `check` goal both in the CI and locally so that we have some common tooling for checking the readiness of the codebase. That's why `check` uses `check_fmt` and not `fmt` - it should fail if the code doesn't conform.

I meant `ready` as an alternative of the `check` goal for everyday work on the code. You don't want to merely check that formatting is OK, you want the formatting to be applied straight away.

sparkoo (approved):
I was thinking to include this to some existing target, because I feel like we have lot of targets already with some of them overlapping functionality. However, I think this one is really useful to run before commit.

Can you please add comment with intention of this target and maybe include it in contributors doc?

mshaposhnik (approved):

PR #369, opened 2022-12-09
Test setup continued

metlos:
### What does this PR do?
This converts the rest of the integration tests to the new Test Setup logic.

https://issues.redhat.com/browse/SVPI-198
https://github.com/redhat-appstudio/service-provider-integration-operator/pull/348

Note that this builds on top of #348 so the diff is rather large now but should be much smaller once #348 is merged.

### How to test this PR?
no new logic, just changes in the integration tests


metlos:
converted back to draft so that we indicate better this needs to wait until #348 is merged.

mshaposhnik (approved):

PR #375, opened 2022-12-15
Prometheus metric that shows time is needed to complete OAuth flow.

skabashnyuk:
### What does this PR do?
Introduces new metric `redhat_appstudio_spi_oauth_flow_complete_time` that is published on spi-oauth service. That is the histogram metric that is monitoring the time needed to complete the OAuth flow. It has labels: type and URL. 

Known limitations.
SPI by itself supports multiple instances of the same provider type at the same time. Such as two or more different GitLab instances. Metrics would be collected correctly because they are distinct by type and service provider URL. The limitation has a Grafana dashboard. It is able only to differentiate data by service provider type.  In case of multiple instances at the same time, the data might be displayed incorrectly.  One of the ways to solve this limitation is to create a dedicated dashboard with variables `type` and `url` like controller's dashboard.

Знімок екрана 2022-12-28 о 14 58 13

Знімок екрана 2022-12-28 о 14 56 32

https://issues.redhat.com/browse/SVPI-106

### How to test this PR?
- deploy new spi oauth image `quay.io/skabashn/service-provider-integration-oauth:metricsoauthflow_2022_12_29__11_22_24`
- create binding for Github, Quay or GitLab. And follow OAuth flow
- `kubectl -n spi-system exec deployment/spi-oauth-service -c oauth -t -- curl --silent localhost:8080/metrics  | sort | grep redhat_appstudio_spi_oauth_flow_complete_time` 
- or deploy monitoring `./hack/deploy-minikube-monitoring.sh`SPI Outbound traffic` dashboard

### DIFF
--- a/api/v1beta1/spiaccesstokenbinding_types.go
+++ b/api/v1beta1/spiaccesstokenbinding_types.go
@@ -26,6 +26,7 @@ type SPIAccessTokenBindingSpec struct {
 	RepoUrl     string      `json:"repoUrl"`
 	Permissions Permissions `json:"permissions,omitempty"`
 	Secret      SecretSpec  `json:"secret"`
+	Lifetime    string      `json:"lifetime,omitempty"`
 }
 
 // SPIAccessTokenBindingStatus defines the observed state of SPIAccessTokenBinding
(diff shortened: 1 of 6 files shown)


oauth/statestorage.go
```diff
@@ -40,19 +41,25 @@ const (
 )
 
 func (s StateStorage) VeilRealState(req *http.Request) (string, error) {
+
 	log := log.FromContext(req.Context())
+	log.Info("VeilRealState1")
```
metlos, line 46: I'd either reformulate these or remove them altogether.

oauth/metrics.go
```diff
@@ -41,3 +50,57 @@ func HttpServiceInstrumentMetricHandler(reg prometheus.Registerer, handler http.
...
+func (c *CompletedFlowMetricHandler) ServeHTTP(w http.ResponseWriter, req *http.Request) {
+	log := log.FromContext(req.Context())
+	log.Info("CompletedFlowMetricHandler ServeHTTP", "Id", c.spConfiguration.ClientId)
+	statusWriter := NewStatusHeaderResponseWriter(w)
+	c.handler.ServeHTTP(statusWriter, req)
+	veiledAt, err := c.stateStorage.StateVeiledAt(req.Context(), req)
+	if err != nil {
+		log.Error(err, "Unknown state")
+	}
+	OAuthFlowCompleteTimeMetric.WithLabelValues(string(c.spConfiguration.ServiceProviderType), c.spConfiguration.ClientId).Observe(time.Since(veiledAt).Seconds())
```
metlos, line 104: Why are you using the `clientId` as the value of the `url` metric label?

metlos (approved):

PR #377, opened 2022-12-16
Configurable lifetime per binding

mshaposhnik:
### What does this PR do?
Allows to set configurable lifetime per every given binding or `-1` for infinite lifetime.

https://issues.redhat.com/browse/SVPI-278

### How to test this PR?

- Create a binding with relatively low `spec.lifetime`, like 70 sec, wait for it do disappear after this time.
- Configure operator to have low defaut timeout, like 20 sec, and create binding with `-1` as a `spec.lifetime`, it should not be deleted after default TTL passing.


controllers/spiaccesstokenbinding_controller.go
```diff
@@ -328,7 +346,12 @@ func (r *SPIAccessTokenBindingReconciler) Reconcile(ctx context.Context, req ctr
 			r.updateBindingStatusError(ctx, &binding, api.SPIAccessTokenBindingErrorReasonTokenSync, err)
 		}
 	}
-	return ctrl.Result{RequeueAfter: r.durationUntilNextReconcile(&binding)}, nil
+	if infiniteDuration {
+		// no need to re-schedule by any timeout
+		return ctrl.Result{}, nil
+	} else {
+		return ctrl.Result{RequeueAfter: time.Until(binding.CreationTimestamp.Add(expectedLifetimeDuration).Add(r.Configuration.DeletionGracePeriod))}, nil
+	}
```
xbaran4, line 354:
```suggestion
	if infiniteDuration {
		// no need to re-schedule by any timeout
		return ctrl.Result{}, nil
	return ctrl.Result{RequeueAfter: time.Until(binding.CreationTimestamp.Add(expectedLifetimeDuration).Add(r.Configuration.DeletionGracePeriod))}, nil
```

controllers/spiaccesstokenbinding_controller.go
```diff
@@ -203,9 +204,28 @@ func (r *SPIAccessTokenBindingReconciler) Reconcile(ctx context.Context, req ctr
 		return ctrl.Result{}, nil
 	}
 
+	var expectedLifetimeDuration = r.Configuration.AccessTokenBindingTtl
+	var infiniteDuration = false
```
xbaran4, line 208:
```suggestion
	expectedLifetimeDuration := r.Configuration.AccessTokenBindingTtl
	infiniteDuration := false
```

xbaran4 (approved):

docs/USER.md
```diff
@@ -270,6 +271,26 @@ Please note that SPI service didn't perform any specific validity checks
 for the user provided credentials, so it's assumed user provides a correct data.
 
 
+## Overriding default binding lifetime
+In general, bindings (and their dependant secrets) are not supposed to be long-lived. Their TTL in configurable globally, 
+and by default it is set to 2 hrs. But some tasks require more long-lived credentials to be available, so the creator of the binding
+may override the default setting by specifying `spec.lifetime` field of the binding.
+It is accepts any standard time periods, like `2h30m`, `90s` etc, except the negative values and values less than
```
metlos: Any standard time period greater than `60s`, like `2h30m` or `90s`, is accepted. To express infinite lifetime, the value of `-1` is also allowed. The bindings with infinite lifetime will only be removed when deleted from the cluster.

controllers/spiaccesstokenbinding_controller.go
```diff
@@ -461,6 +478,7 @@ func (r *SPIAccessTokenBindingReconciler) persistWithMatchingLabels(ctx context.
 
 // updateBindingStatusError updates the status of the binding with the provided error
 func (r *SPIAccessTokenBindingReconciler) updateBindingStatusError(ctx context.Context, binding *api.SPIAccessTokenBinding, reason api.SPIAccessTokenBindingErrorReason, err error) {
+	binding.Status.Phase = api.SPIAccessTokenBindingPhaseError
```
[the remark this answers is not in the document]
mshaposhnik: ok i see, i'll return back

controllers/spiaccesstokenbinding_controller.go
```diff
@@ -203,9 +204,30 @@ func (r *SPIAccessTokenBindingReconciler) Reconcile(ctx context.Context, req ctr
 		return ctrl.Result{}, nil
 	}
 
+	expectedLifetimeDuration := r.Configuration.AccessTokenBindingTtl
+	infiniteDuration := false
+	if binding.Spec.Lifetime != "" {
```
sparkoo, line 209: can we maybe instead of this another block of ifs have single function (with proper tests) like `func tokenLifetim(...) (*time.Duration, error)` that will take all this logic and return final duration of the binding or nil for infinite?

mshaposhnik: well. if we don't afraid the one-time funcs and `nil` returns (oh gave me those optionals back), that probably looks little better.

docs/USER.md
```diff
@@ -270,6 +271,26 @@ Please note that SPI service didn't perform any specific validity checks
 for the user provided credentials, so it's assumed user provides a correct data.
 
 
+## Overriding default binding lifetime
+In general, bindings (and their dependent secrets) are not supposed to be long-lived. Their TTL is configured globally, 
+and by defaults to 2 hrs. Some tasks may require the credentials to be available for a longer time, so the creator of the binding
```
metlos, line 276: ... and defaults to 2 hrs.
or
... and is 2 hrs by default.

docs/USER.md
```diff
@@ -270,6 +271,26 @@ Please note that SPI service didn't perform any specific validity checks
...
+may override the default setting by specifying `spec.lifetime` field of the binding.
+It accepts any standard time periods, like `2h30m`, `90s` etc, except the negative values and values less than
```
metlos, line 278: IMHO, this sentence and the sentence after it are overly complicated.

docs/USER.md
mshaposhnik: fixed

docs/USER.md
mshaposhnik: tried to simplify a bit

controllers/spiaccesstokenbinding_controller.go
```diff
@@ -328,7 +336,31 @@ func (r *SPIAccessTokenBindingReconciler) Reconcile(ctx context.Context, req ctr
...
+	if expectedLifetimeDuration == nil {
+		// no need to re-schedule by any timeout
+		return ctrl.Result{}, nil
+	}
+
+	return ctrl.Result{RequeueAfter: time.Until(binding.CreationTimestamp.Add(*expectedLifetimeDuration).Add(r.Configuration.DeletionGracePeriod))}, nil
+}
+
+// returns user specified binding lifetime or default binding lifetime or nil if set to infinite
+func bindingLifetime(r *SPIAccessTokenBindingReconciler, binding api.SPIAccessTokenBinding) (*time.Duration, error) {
```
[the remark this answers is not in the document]
mshaposhnik: ok

controllers/spiaccesstokenbinding_controller.go
```diff
@@ -328,7 +336,31 @@ func (r *SPIAccessTokenBindingReconciler) Reconcile(ctx context.Context, req ctr
 			r.updateBindingStatusError(ctx, &binding, api.SPIAccessTokenBindingErrorReasonTokenSync, err)
 		}
 	}
-	return ctrl.Result{RequeueAfter: r.durationUntilNextReconcile(&binding)}, nil
+	if expectedLifetimeDuration == nil {
+		// no need to re-schedule by any timeout
```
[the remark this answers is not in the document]
mshaposhnik: i afraid it may affect the logic

PR #387, opened 2023-01-04
Handle situation when the token is not found in storage

skabashnyuk:
### What does this PR do?
Correctly handling situations when the token is not found to avoid `invalid memory address or nil pointer dereference` 

n/a

https://issues.redhat.com/browse/SVPI-289

### How to test this PR?
n/a

### DIFF
--- a/pkg/serviceprovider/github/githubclientbuilder.go
+++ b/pkg/serviceprovider/github/githubclientbuilder.go
@@ -44,7 +44,7 @@ func (g *githubClientBuilder) createAuthenticatedGhClient(ctx context.Context, s
 		return nil, fmt.Errorf("failed to get token from storage for %s/%s: %w", spiToken.Namespace, spiToken.Name, tsErr)
 	}
 	if tokenData == nil {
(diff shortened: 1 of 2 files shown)


pkg/serviceprovider/github/githubclientbuilder.go
```diff
@@ -40,6 +43,10 @@ func (g *githubClientBuilder) createAuthenticatedGhClient(ctx context.Context, s
 		lg.Error(tsErr, "failed to get token from storage for", "token", spiToken)
 		return nil, fmt.Errorf("failed to get token from storage for %s/%s: %w", spiToken.Namespace, spiToken.Name, tsErr)
 	}
+	if tokenData == nil {
+		lg.Error(accessTokenNotFoundError, "token data not found", "token-name", spiToken.Name, "tokenData", tokenData)
```
xbaran4, line 47: I understand that the tokenData here is always logged as "nil" but it scares me that we might expose secret data if we accidentally remove the if statement (maybe I'm paranoid :)). Plus, if the data is always logged as "nil" is it needed to log this value at all?

xbaran4 (approved):

PR #392, opened 2023-01-05
move service provider default values into central place

sparkoo:
### What does this PR do?
refactor default values for all supported service providers into single central place

helps with:
 - https://issues.redhat.com/browse/SVPI-270
 - https://issues.redhat.com/browse/SVPI-189

### How to test this PR?

images:
 - `quay.io/mvala/service-provider-integration-oauth:svpi270-defaults`
 - `quay.io/mvala/service-provider-integration-operator:svpi270-defaults`

it's refactoring, so functionality should stay the same. Test whatever worked and it should work the same.


pkg/spi-shared/config/defaults.go
```diff
@@ -0,0 +1,50 @@
...
+
+package config
+
+import (
+	"golang.org/x/oauth2"
+	"golang.org/x/oauth2/github"
+)
+
+// ServiceProviderDefaults configuration containing default values used to initialize supported service providers
+type ServiceProviderDefaults struct {
```
metlos, line 23: what if we made `ServiceProviderType` a struct (with `Name` being the current value of the different types and additional fields as in this defaults struct)?

pkg/spi-shared/config/defaults.go
```diff
@@ -0,0 +1,50 @@
...
+	SpType   ServiceProviderType
+	Endpoint oauth2.Endpoint
+	UrlHost  string // default host of service provider. ex.: `github.com`
+	BaseUrl  string // default base url of service provider, typically scheme+host. ex: `https://github.com`
+}
+
+// all servise provider types we support, including default values
+var SupportedServiceProvidersDefaults []ServiceProviderDefaults = []ServiceProviderDefaults{
```
metlos, line 31: If we used `ServiceProviderType` as the holder of the defaults, this could be called `SupportedServiceProviderTypes` which IMHO reads ever so slightly better.

Also, I would actually make the individual default or type instances public (and only reference them in this array) so that we could avoid the naming explosion for the constants specifying the default values of each of the fields.

E.g. instead of `config.GithubSaasHost` and `config.QuaySaasHost` and `config.GitlabSaasHost`, we'd have `config.GithubServiceProviderDefaults.SaasHost`, `config.QuayServiceProviderDefaults.SaasHost` and `config.GitlabServiceProviderDefaults.SaasHost`. IMHO, that is slightly more consistent and provides less room for typos and overall cognitive overload.

pkg/spi-shared/config/github.go
```diff
@@ -11,6 +11,11 @@
...
 // limitations under the License.
 
-package oauth
+package config
 
-var GithubSaasHost = "github.com"
+var ServiceProviderTypeGitHub ServiceProviderType = "GitHub"
+
+var (
+	GithubSaasHost    = "github.com"
```
metlos, line 19: I might be unnecessarily purist but I'd not make these public but rather "hide" them in the defaults or (if you accept my suggestion) service provider type instances.

E.g. instead of using this as `config.GithubSaasHost`, you'd `config.GithubServiceProviderDefaults.SaasHost` or `config.GithubServiceProviderType.SaasHost`.

pkg/spi-shared/config/defaults.go
```diff
@@ -0,0 +1,50 @@
...
+	"golang.org/x/oauth2"
+	"golang.org/x/oauth2/github"
+)
+
+// ServiceProviderDefaults configuration containing default values used to initialize supported service providers
+type ServiceProviderDefaults struct {
+	SpType   ServiceProviderType
+	Endpoint oauth2.Endpoint
+	UrlHost  string // default host of service provider. ex.: `github.com`
+	BaseUrl  string // default base url of service provider, typically scheme+host. ex: `https://github.com`
```
metlos, line 27: The difference between `BaseUrl` and `UrlHost` seems kinda cosmetic. Is there a way of deriving one from another? Or do we want to support the base url being on a subpath of the host or something?

pkg/spi-shared/config/defaults.go
```diff
@@ -0,0 +1,50 @@
...
+
+import (
+	"golang.org/x/oauth2"
+	"golang.org/x/oauth2/github"
+)
+
+// ServiceProviderDefaults configuration containing default values used to initialize supported service providers
+type ServiceProviderDefaults struct {
+	SpType   ServiceProviderType
+	Endpoint oauth2.Endpoint
```
metlos, line 25: `OAuthEndpoint` to be more precise?

metlos (commented):
I think this is a step in the right direction, but I would like to propose one additional change (apart from the suggestions below):

Change the `KnownInitializers()` function to `GetInitializer(ServiceProviderType)` so that, API-wise, there is just one place where we need to list all the service providers we know about?

pkg/spi-shared/config/defaults.go
sparkoo: only reason I've put both here is so we don't have to type `"https://"+host` everywhere where we need it. I'll check the usages and let's see if it make sense.

sparkoo:
> Change the `KnownInitializers()` function to `GetInitializer(ServiceProviderType)` so that, API-wise, there is just one place where we need to list all the service providers we know about?

I wanted to do that, but at the same time I wanted to keep `config` package to not have any internal dependencies. `Initializer` here is quite heavy dependency.

pkg/spi-shared/config/config.go
```diff
@@ -14,22 +14,38 @@
...
+	DefaultHost          string          // default host of the service provider. ex.: `github.com`
+	DefaultBaseUrl       string          // default base url of the service provider, typically scheme+host. ex: `https://github.com`
+}
 
-const (
-	ServiceProviderTypeGitHub          ServiceProviderType = "GitHub"
-	ServiceProviderTypeQuay            ServiceProviderType = "Quay"
-	ServiceProviderTypeGitLab          ServiceProviderType = "GitLab"
-	ServiceProviderTypeHostCredentials ServiceProviderType = "HostCredentials"
+// all servise provider types we support, including default values
```
metlos, line 35: ... all service provider types ...

pkg/spi-shared/config/config.go
```diff
@@ -14,22 +14,38 @@
...
+var SupportedServiceProviderTypes []ServiceProviderType = []ServiceProviderType{
+	ServiceProviderTypeGitHub,
+	ServiceProviderTypeGitLab,
+	ServiceProviderTypeQuay,
```
metlos, line 39: Why is it important to leave out the host creds provider from this list? Maybe it'd be worth mentioning that fact in some kind of comment here...

pkg/spi-shared/config/config.go
```diff
@@ -57,7 +73,28 @@ type CommonCliArgs struct {
...
+// ServiceProviderConfiguration contains configuration for a single service provider configured with the SPI. This
+// mainly contains config.yaml of the OAuth application within the service provider.
+type persistedServiceProviderConfiguration struct {
+	// ClientId is the client ID of the OAuth application that the SPI uses to access the service provider.
+	ClientId string `yaml:"clientId"`
+
+	// ClientSecret is the client secret of the OAuth application that the SPI uses to access the service provider.
+	ClientSecret string `yaml:"clientSecret"`
+
+	// ServiceProviderType is the type of the service provider. This must be one of the supported values: GitHub, Quay
```
metlos, line 88: the name in the doc comment no longer corresponds to the name of the field. Also, Gitlab is not mentioned as a possible value.

pkg/spi-shared/config/config.go
```diff
@@ -57,7 +73,28 @@ type CommonCliArgs struct {
...
+	ServiceProviderName ServiceProviderName `yaml:"type"`
+
+	// ServiceProviderBaseUrl is the base URL of the service provider. This can be omitted for certain service provider
+	// types, like GitHub that only can have 1 well-known base URL.
+	ServiceProviderBaseUrl string `yaml:"baseUrl,omitempty"`
+
+	// Extra is the extra configuration required for some service providers to be able to uniquely identify them. E.g.
+	// for Quay, we require to know the organization for which the OAuth application is defined for.
```
metlos, line 96: The mention about Quay is incorrect, let's just delete it from the comment when we're at it? 😃

pkg/spi-shared/config/config.go
```diff
@@ -75,29 +112,50 @@ type SharedConfiguration struct {
...
-	ServiceProviderType ServiceProviderType `yaml:"type"`
+	ServiceProviderType ServiceProviderType
 
 	// ServiceProviderBaseUrl is the base URL of the service provider. This can be omitted for certain service provider
 	// types, like GitHub that only can have 1 well-known base URL.
-	ServiceProviderBaseUrl string `yaml:"baseUrl,omitempty"`
+	ServiceProviderBaseUrl string
 
 	// Extra is the extra configuration required for some service providers to be able to uniquely identify them. E.g.
 	// for Quay, we require to know the organization for which the OAuth application is defined for.
```
metlos, line 128: Could you please remove this incorrect comment about quay?

pkg/spi-shared/config/config.go
```diff
@@ -75,29 +112,50 @@ type SharedConfiguration struct {
...
-	Extra map[string]string `yaml:"extra,omitempty"`
+	Extra map[string]string
+}
+
+func init() {
+	// It's handy to have serviceproviders reachable by it's name. Let's fill in such map from central list 'SupportedServiceProviderTypes'.
```
metlos, line 133: ... by their names.

metlos (commented):
Look awesome apart from the minor comments about comments below 😃

pkg/spi-shared/config/config.go
sparkoo: fixed

pkg/spi-shared/config/config.go
sparkoo: fixed

pkg/spi-shared/config/config.go
sparkoo: fixed

pkg/spi-shared/config/config.go
sparkoo: fixed

pkg/spi-shared/config/config.go
sparkoo: fixed

oauth/common.go
```diff
@@ -97,7 +97,7 @@ func (c *commonController) Authenticate(w http.ResponseWriter, r *http.Request,
 		LogDebugAndWriteResponse(ctx, w, http.StatusUnauthorized, "authenticating the request in Kubernetes unsuccessful")
 		return
 	}
-	AuditLogWithTokenInfo(ctx, "OAuth authentication flow started", state.TokenNamespace, state.TokenName, "scopes", state.Scopes, "providerType", string(state.ServiceProviderType), "providerUrl", state.ServiceProviderUrl)
+	AuditLogWithTokenInfo(ctx, "OAuth authentication flow started", state.TokenNamespace, state.TokenName, "scopes", state.Scopes, "providerType", state.ServiceProviderName, "providerUrl", state.ServiceProviderUrl)
```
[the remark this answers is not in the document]
sparkoo: fixed

oauth/common.go
```diff
@@ -144,7 +144,7 @@ func (c *commonController) Callback(ctx context.Context, w http.ResponseWriter,
 		LogErrorAndWriteResponse(ctx, w, http.StatusInternalServerError, "failed to store token data to cluster", err)
 		return
 	}
-	AuditLogWithTokenInfo(ctx, "OAuth authentication completed successfully", exchange.TokenNamespace, exchange.TokenName, "scopes", exchange.Scopes, "providerType", string(exchange.ServiceProviderType), "providerUrl", exchange.ServiceProviderUrl)
+	AuditLogWithTokenInfo(ctx, "OAuth authentication completed successfully", exchange.TokenNamespace, exchange.TokenName, "scopes", exchange.Scopes, "providerType", exchange.ServiceProviderName, "providerUrl", exchange.ServiceProviderUrl)
```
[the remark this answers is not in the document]
sparkoo: fixed

pkg/spi-shared/config/config.go
sparkoo: fixed

pkg/spi-shared/config/defaults.go
sparkoo: It has quite some usages. I keep it there if you don't mind.

PR #396, opened 2023-01-10
fix SVPI-232

gazarenkov:
### What does this PR do?
Allowing not to put secret.type or put "Opaque".
In this case SPIAccessTokenBinding will create Opaque secret with "token" field filled with token data 

n/a

https://issues.redhat.com/projects/SVPI/issues/SVPI-232

###  How to test this PR ?

- Create a SPIAccessTokenBinding with spec.secret.type : Opaque OR without this field
- Make sure it's status.phase == Injected (may require uploading Access Token with OAuth or manually)
- Check the Secret by name mentioned in status.syncedObjectRef.name. It should have "token" key with non empty value (token data)


pkg/serviceprovider/access_token_mapper.go
```diff
@@ -63,6 +63,8 @@ func (at AccessTokenMapper) ToSecretType(secretType corev1.SecretType) map[strin
 			at.Token)
 	case corev1.SecretTypeSSHAuth:
 		ret[corev1.SSHAuthPrivateKey] = at.Token
+	default:
+		ret["token"] = at.Token
```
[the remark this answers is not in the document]
gazarenkov: +

mshaposhnik (approved):

pkg/serviceprovider/access_token_mapper.go
```diff
@@ -63,6 +66,8 @@ func (at AccessTokenMapper) ToSecretType(secretType corev1.SecretType) map[strin
 			at.Token)
 	case corev1.SecretTypeSSHAuth:
 		ret[corev1.SSHAuthPrivateKey] = at.Token
+	default:
+		ret[tokenKey] = at.Token
```
metlos, line 70: This means that the `Opaque` secrets will ALWAYS have a field called `token`, even if there is a field mapping on the binding that would be configured differently. Are we OK with that?

metlos (commented):
The curre

pkg/serviceprovider/access_token_mapper.go
gazarenkov: @metlos Do we have a mechanism to map it?

gazarenkov: @skabashnyuk I would like to understand if the double-token-field ("token" and mapped one) is the problem of Opaque token only or it may happen with any Secret type. 
For example what if we have mapping for token with SA type?
Will we see 
extra : <token>
AND
mapped-name: <token>
?

gazarenkov: done

pkg/serviceprovider/access_token_mapper.go
```diff
@@ -63,14 +66,24 @@ func (at AccessTokenMapper) ToSecretType(secretType corev1.SecretType) map[strin
...
 }
 
-// FillByMapping sets the data from the mapper into the provided map according to the settings specified in the provided
+// fillByMapping sets the data from the mapper into the provided map according to the settings specified in the provided
 // mapping.
-func (at AccessTokenMapper) FillByMapping(mapping *api.TokenFieldMapping, existingMap map[string]string) {
+func (at AccessTokenMapper) fillByMapping(mapping *api.TokenFieldMapping, existingMap map[string]string) {
+
+	// if there are no mapping.Token - add "token" key
+	if mapping.Token == "" {
```
metlos, line 81: I would only add the `token` "default mapping" if there is no other mapping. I.e. the user may choose not to put the token into the secret and we should not disallow that IMHO. Here, I'd just add an `IsEmpty()` method on the TokenFieldMapping checking that all the fields are empty and use that in this method.

```go
if mapping.IsEmpty() {
  existingMap[tokenKey] = at.Token
  return
}

// the rest of the method unchanged from before this PR
```

gazarenkov: @metlos is there any sense in secret which does not contain token? 
I guessed that's the main reason of this issue to not to make it happen.

metlos (approved):

PR #407, opened 2023-01-19
Refactor contoller to be more idempotent

mshaposhnik:
Signed-off-by: Max Shaposhnyk <mshaposh@redhat.com>

### What does this PR do?
This PR changes the way of `SPIFileContentRequest` reconcillation in the more status-independent way.
It shifts the linking between File request and binding from the status field in the file request (updates to which may be not reflected/seen in the reconcillations that was scheduled on some intermediate resource versions of the request CR, and cause to re-wtiting status concurrently) to the specific label in the binding. 

That also having a positive effect in more precise bindings watching, avoiding processing the events from bindings which are unrelated to the file requests.

https://issues.redhat.com/browse/SVPI-290

### How to test this PR?

The only difference with current main is that duplicate bidndings for the same FileRequest should never appear in tests nor in live cluster anymore. (reviewers, please run `make itest focus=SPIFileContentRequest`  several times)


controllers/spifilecontentrequest_controller.go
```diff
@@ -154,50 +133,53 @@ func (r *SPIFileContentRequestReconciler) Reconcile(ctx context.Context, req ctr
...
+			log.FromContext(ctx).Error(err, "failed to update the file binding link", "error", err)
+			return reconcile.Result{Requeue: true}, nil
 		}
+		return ctrl.Result{Requeue: true}, nil //TODO: better to requeue after some interval
+	} else {
 		//binding not yet fully ready to work with, let's try next time
-		if binding.Status.Phase == "" || binding.Status.UploadUrl == "" {
-			return reconcile.Result{}, nil
+		if associatedBinding.Status.Phase == "" || associatedBinding.Status.UploadUrl == "" {
+			return reconcile.Result{Requeue: true}, nil
```
metlos, line 172: This creates a sort of busy loop until the associated binding is reconciled. Wouldn't it be better to reintroduce the watch on the bindings (and maybe rework it with the newly introduced label)?

controllers/spifilecontentrequest_controller.go
```diff
@@ -154,50 +133,53 @@ func (r *SPIFileContentRequestReconciler) Reconcile(ctx context.Context, req ctr
...
-	if request.Status.Phase == "" {
-		// check if the URL is processable, otherwise fail fast
-		sp, _ := r.ServiceProviderFactory.FromRepoUrl(ctx, request.Spec.RepoUrl, request.Namespace)
-		if sp.GetDownloadFileCapability() == nil {
-			r.updateFileRequestStatusError(ctx, &request, serviceprovider.FileDownloadNotSupportedError{})
-			return ctrl.Result{}, serviceprovider.FileDownloadNotSupportedError{}
-		}
-		request.Status.Phase = api.SPIFileContentRequestPhaseAwaitingBinding
-	} else if request.Status.Phase == api.SPIFileContentRequestPhaseDelivered {
+	if request.Status.Phase == api.SPIFileContentRequestPhaseDelivered {
```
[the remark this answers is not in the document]
mshaposhnik: yep that's by design

xbaran4 (approved):

controllers/spifilecontentrequest_controller.go
```diff
@@ -74,28 +95,18 @@ func (r *SPIFileContentRequestReconciler) SetupWithManager(mgr ctrl.Manager) err
...
-				if fr.Status.LinkedBindingName == o.GetName() {
-					ret = append(ret, reconcile.Request{
-						NamespacedName: types.NamespacedName{
-							Name:      fr.Name,
-							Namespace: fr.Namespace,
-						},
-					})
-				}
-			}
+			ret := make([]reconcile.Request, 0, 1)
```
sparkoo, line 101: I don't understand why this construct? why first make 0 slice with capped size to 1 then immediatelly append and return? why not:
```
return []reconcile.Request{
reconcile.Request{
				NamespacedName: types.NamespacedName{
					Name:      o.GetLabels()[LinkedFileRequestLabel],
					Namespace: o.GetNamespace(),
				},
			}
}
```
?

controllers/spifilecontentrequest_controller.go
```diff
@@ -154,88 +165,101 @@ func (r *SPIFileContentRequestReconciler) Reconcile(ctx context.Context, req ctr
...
-			return ctrl.Result{}, serviceprovider.FileDownloadNotSupportedError{}
-		}
-		request.Status.Phase = api.SPIFileContentRequestPhaseAwaitingBinding
-	} else if request.Status.Phase == api.SPIFileContentRequestPhaseDelivered {
+	if request.Status.Phase == api.SPIFileContentRequestPhaseDelivered {
 		// already injected, nothing to do
 		return ctrl.Result{}, nil
 	}
+	// find associated binding
+	var associatedBinding *api.SPIAccessTokenBinding = nil
```
sparkoo, line 173: I don't think you have to explicitly set it to `nil`, but nothing wrong with that

sparkoo (approved):
I'd like to see some stuff extracted to functions and tested. These very long methods full of ifs are hard to read. otherwise lgtm

PR #408, opened 2023-01-19
Add the ability to associate the secret generate by the binding with a potentially pre-existing service account.

metlos:
### What does this PR do?
`$TITLE`

https://issues.redhat.com/browse/SVPI-233

### How to test this PR?

Service account token:

1. create an empty service account:
   ```yaml
   apiVersion: v1
   kind: ServiceAccount
   metadata:
     name: test-sa
     namespace: default
   ```
2. create a binding with a service account definition:
    ```yaml
    ...
    spec:
       secret:
         type: kubernetes.io/service-account-token
         annotations:
           "kubernetes.io/service-account.name": "test-sa"
         linkedTo:
         - serviceAccount:
             reference:
               name: test-sa
    ```
3. upload the token data or go through the OAuth flow
4. check that the binding is injected and mentions the `serviceAccountNames` in its status.
5. check that the secret has the correct type and has the prescribed annotation.
6. check that the newly created service account mentions the secret in its `secrets` field.
7. check that the secret has the token in the `extra` field and `ca.crt`, `namespace` and `token` auto-filled by the Kubernetes control plane.
8. delete the binding
9. check that the secret has been deleted.
10. check that the service account is still there

(we do not support managed service account token secrets. These tokens have very limited utility in SPI, because all we can do is inject extra fields into them, while the main work is done by the Kubernetes infrastructure (injection of the actual service account token and certificate)).
 
Image pull secret:
1. Create the binding with:
   ```yaml
    ...
    spec:
       secret:
         type: kubernetes.io/dockerconfigjson
         linkedTo:
         - serviceAccount:
             as: imagePullSecret
             managed:
               generateName: test-sa-
    ```
3. upload the token data or go through the OAuth flow
6. check that the newly created service account mentions the secret in its `imagePullSecrets` field.
2. the service account should link the secret in its `imagePullSecrets` section.
8. delete the binding
9. check that both the secret and the service account have been deleted.

Other secret types:

It should be possible to link other secret types with the service account.

1. Repeat the steps for the image pull secrets but replace the secret type for `kubernetes.io/basic-auth` and remove the `as: imagePullSecret` definition.
5. check that the secret appears in the list of the secrets in the service account.
6. the secret should NOT have any new annotation (as opposed to the service account token secret)
7. no auto-filled stuff should be present in the secret.

Unmanaged:
1. Unmanaged means that the service account must already exist in the cluster, so let's just use the `default` SA.
1. Link some secret to the `default` SA manually (i.e. add a reference to the secret to the `secrets` field of the `default` SA).
1. Repeat the above steps for any secret type but replace the 
   ```yaml
   managed:
     generateName: ...
   ``` 
   with
   ```yaml
   reference:
      name: default
   ```
1. check that the binding secret is merely added to the list of the service acount secrets, not overwriting any pre-existing links.
1. The service account should stay there after the binding is deleted.

## Additional notes
There is an image with the latest code from this PR @ `quay.io/lkrejci/service-provider-integration-oauth:svpi-233`


api/v1beta1/spiaccesstokenbinding_types.go
```diff
@@ -146,3 +188,24 @@ func (in *SPIAccessTokenBinding) ObjNamespace() string {
...
+	}
+
+	saAnnotation := in.Spec.Secret.Annotations[corev1.ServiceAccountNameKey]
+
+	if in.Spec.Secret.Type == corev1.SecretTypeServiceAccountToken && saAnnotation == "" && in.Spec.ServiceAccount == nil {
+		ret.Consistency = append(ret.Consistency, fmt.Sprintf("the secret is configured with \"%s\" type, but \"%s\" annotation is configured neither in the definition of the binding secret, nor is a service account configured in the binding. This is invalid.", corev1.SecretTypeServiceAccountToken, corev1.ServiceAccountNameKey))
+	}
+
+	if in.Spec.Secret.Type == corev1.SecretTypeServiceAccountToken && saAnnotation != "" && in.Spec.ServiceAccount != nil && in.Spec.ServiceAccount.Name != saAnnotation {
+		ret.Consistency = append(ret.Consistency, "both the service account name annotation on the secret and an explicit service account are configured but differ in valuei. Either define just one of them or make sure the values match.")
```
mshaposhnik, line 207: typo in `valuei`

mshaposhnik (approved):

controllers/spiaccesstokenbinding_controller.go
```diff
@@ -308,19 +356,46 @@ func (r *SPIAccessTokenBindingReconciler) Reconcile(ctx context.Context, req ctr
...
+	serviceAccountName := binding.Status.ServiceAccountName
+
 	if token.Status.Phase == api.SPIAccessTokenPhaseReady {
-		ref, err := r.syncSecret(ctx, sp, &binding, token)
+		// linking the service account with a secret is a 3-step process.
+		// First, an empty service account needs to be created.
+		// Second, a secret linking to the service account needs to be created.
+		// Third, the service account needs to be updated with the link to the secret.
+
+		// TODO this can leave the SA behind if we subsequently fail to update the status of the binding
```
sparkoo, line 368: is issue created for this TODO?

controllers/spiaccesstokenbinding_controller.go
```diff
@@ -308,19 +356,46 @@ func (r *SPIAccessTokenBindingReconciler) Reconcile(ctx context.Context, req ctr
...
+		if err != nil {
+			lg.Error(err, "unable to sync the service account")
+			return ctrl.Result{}, fmt.Errorf("failed to sync the service account: %w", err)
+		}
+		if serviceAccount != nil {
+			binding.Status.ServiceAccountName = serviceAccount.Name
+			lg.Info("service account linked", "saName", serviceAccount.Name, "managed", binding.Spec.ServiceAccount.Managed)
+		}
+
+		// TODO this can leave the secret behind if we subsequently fail to update the status of the binding
```
sparkoo, line 379: is issue created for this TODO?

controllers/spiaccesstokenbinding_controller.go
```diff
@@ -344,6 +428,24 @@ func (r *SPIAccessTokenBindingReconciler) Reconcile(ctx context.Context, req ctr
 	return ctrl.Result{RequeueAfter: time.Until(binding.CreationTimestamp.Add(*expectedLifetimeDuration).Add(r.Configuration.DeletionGracePeriod))}, nil
 }
 
+func (r *SPIAccessTokenBindingReconciler) migrateFinalizers(ctx context.Context, binding *api.SPIAccessTokenBinding) error {
```
sparkoo, line 431: what does it mean to migrate finalizers?

controllers/spiaccesstokenbinding_controller.go
```diff
@@ -564,27 +677,141 @@ func (r *SPIAccessTokenBindingReconciler) syncSecret(ctx context.Context, sp ser
...
 	if err != nil {
 		r.updateBindingStatusError(ctx, binding, api.SPIAccessTokenBindingErrorReasonTokenSync, err)
-		return api.TargetObjectRef{}, fmt.Errorf("failed to sync the secret with the token data: %w", err)
+		return nil, fmt.Errorf("failed to sync the secret with the token data: %w", err)
 	}
-	return toObjectRef(obj), nil
+	return obj.(*corev1.Secret), nil
+}
+
+func (r *SPIAccessTokenBindingReconciler) getOrCreateServiceAccount(ctx context.Context, binding *api.SPIAccessTokenBinding) (*corev1.ServiceAccount, error) {
```
sparkoo, line 685: doc and test please

controllers/spiaccesstokenbinding_controller.go
```diff
@@ -564,27 +677,141 @@ func (r *SPIAccessTokenBindingReconciler) syncSecret(ctx context.Context, sp ser
...
+
+	if err != nil {
+		r.updateBindingStatusError(ctx, binding, api.SPIAccessTokenBindingErrorReasonTokenSync, err)
+		return nil, fmt.Errorf("failed to sync the configured service account of the binding: %w", err)
+	}
+
+	return sa, nil
+}
+
+func (r *SPIAccessTokenBindingReconciler) linkSecretToServiceAccount(ctx context.Context, secret *corev1.Secret, sa *corev1.ServiceAccount) error {
```
sparkoo, line 749: doc and test please

sparkoo (commented):
the Reconcile method is getting ridiculously big and complex and this is not helping it. Can you please somehow separate new parts into the tested functions/units, so it's somehow clear what it suppose to be doing in what conditions and so we can feel safer to move code around without breaking everything?

sparkoo:
it's user facing feature. Can you please mention it in docs?

controllers/spiaccesstokenbinding_controller.go
metlos: https://issues.redhat.com/browse/SVPI-210

controllers/spiaccesstokenbinding_controller.go
metlos: This has 100% complete code coverage of the happy path by the integration tests and has a single error handler branch. Not sure what more you want to test.

controllers/spiaccesstokenbinding_controller.go
metlos: This has 100% code coverage in the integration tests.

metlos:
> it's user facing feature. Can you please mention it in docs?

thanks for pointing this out. I've added a couple of examples to `USER.md`.

docs/USER.md
```diff
@@ -156,9 +158,82 @@ type: Opaque
 stringData:
   tokenData: $AT_DATA
 ```
-After reconcillation SPIAccessToken should be filled with the Access Token metadata, it's `status.Phase` should be `Injected` and upload Secret is removed.
+After reconciliation SPIAccessToken should be filled with the Access Token metadata, it's `status.Phase` should be `Injected` and upload Secret is removed.
 In a case if something goes wrong the reason will be written to K8s Event, you can check it with `kubectl get event $upload-secret`.
 
+## Providing secrets to a service account
+
```
gazarenkov, line 165: Would not it be more clear for user to have two different fields (to put one of it), like:
- ServiceAccount: to define and create SA, always managed
- LinkedServiceAccount: with a reference to existed SA (+ linkSecretAs etc)
?

xbaran4: This could mean that users might think that both can be used at the same time.

controllers/spiaccesstokenbinding_controller.go
```diff
@@ -113,16 +162,35 @@ func (r *SPIAccessTokenBindingReconciler) SetupWithManager(mgr ctrl.Manager) err
...
+				enqueueLog.Error(err, "failed to list SPIAccessTokenBindings while determining the ones linked to Secret",
 					"SecretName", o.GetName(), "SecretNamespace", o.GetNamespace())
 				return []reconcile.Request{}
 			}
+
+			logReconciliationRequests(requests, "SPIAccessTokenBinding", o, "Secret")
+
 			return requests
 		})).
+		Watches(&source.Kind{Type: &corev1.ServiceAccount{}}, handler.EnqueueRequestsFromMapFunc(func(o client.Object) []reconcile.Request {
```
sparkoo, line 174: watch `ServiceAccounts` is another heavy object in kubernetes deployments. We already watch secrets and most probably it's cause of memory spikes after startup. Have you seen any substantial increase of memory consumption or spike on startup comparing to main branch? (ideally tested on OpenShift as it is more heavyveight ootb)

controllers/enqueue.go
```diff
@@ -0,0 +1,34 @@
...
+import (
+	"github.com/redhat-appstudio/service-provider-integration-operator/pkg/logs"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/log"
+)
+
+var enqueueLog = log.Log.WithName("enqueue-logger")
+
+func logReconciliationRequests(requests []ctrl.Request, requestKind string, cause client.Object, objectKind string) {
```
sparkoo, line 26: `enqueue.go` sounds quite important. What is responsibility of this file? Any plans to have anything else here than "just" logging?

controllers/spiaccesstokenbinding_controller.go
```diff
@@ -565,30 +724,202 @@ func (r *SPIAccessTokenBindingReconciler) syncSecret(ctx context.Context, sp ser
...
-	return toObjectRef(obj), nil
+	return obj.(*corev1.Secret), nil
 }
 
-func deleteSyncedSecret(ctx context.Context, cl client.Client, secretName string, secretNamespace string) error {
-	if secretName == "" {
-		return nil
+// ensureServiceAccount loads the service account configured in the binding from the cluster or creates a new one if needed.
+// It also makes sure that the service account is correctly labeled.
+func (r *SPIAccessTokenBindingReconciler) ensureServiceAccount(ctx context.Context, binding *api.SPIAccessTokenBinding) (*corev1.ServiceAccount, error) {
```
sparkoo, line 743: would be great to have this function covered with unit tests

controllers/spiaccesstokenbinding_controller.go
```diff
@@ -565,30 +724,202 @@ func (r *SPIAccessTokenBindingReconciler) syncSecret(ctx context.Context, sp ser
...
+	if err != nil {
+		r.updateBindingStatusError(ctx, binding, api.SPIAccessTokenBindingErrorReasonTokenSync, err)
+		return nil, fmt.Errorf("failed to sync the configured service account of the binding: %w", err)
+	}
+
+	return sa, nil
+}
+
+// linkSecretToServiceAccount links the secret with the service account according to the type of the secret.
+func (r *SPIAccessTokenBindingReconciler) linkSecretToServiceAccount(ctx context.Context, linkType api.SecretLinkType, secret *corev1.Secret, sa *corev1.ServiceAccount) error {
```
sparkoo, line 808: would be great to have this function covered with unit tests

controllers/spiaccesstokenbinding_controller.go
```diff
@@ -565,30 +724,202 @@ func (r *SPIAccessTokenBindingReconciler) syncSecret(ctx context.Context, sp ser
...
+			return fmt.Errorf("failed to update the service account with the link to the secret: %w", err)
+		}
+	}
+
+	return nil
+}
+
+// unlinkSecretFromServiceAccount removes the provided secret from any links in the service account (either secrets or image pull secrets fields).
+// Returns `true` if the service account object was changed, `false` otherwise. This does not update the object in the cluster!
+func unlinkSecretFromServiceAccount(secret *corev1.Secret, serviceAccount *corev1.ServiceAccount) bool {
```
sparkoo, line 850: would be great to have this function covered with unit tests

docs/USER.md
gazarenkov: @xbaran4 well, we need to explain this in doc, thets exactly my point, I think user have much more chances to be confused if we have it under the same field.

controllers/spiaccesstokenbinding_controller.go
```diff
@@ -54,23 +56,67 @@ import (
...
 	}
 
+	// the service account secrets are treated specially by Kubernetes that automatically adds "ca.crt", "namespace" and
+	// "token" entries into the secret's data.
+	serviceAccountSecretDiffOpts = cmp.Options{
+		cmpopts.IgnoreFields(corev1.Secret{}, "TypeMeta", "ObjectMeta"),
+		cmp.FilterPath(func(p cmp.Path) bool {
+			return p.Last().String() == ".Data"
+		}, cmp.Comparer(func(a map[string][]byte, b map[string][]byte) bool {
+			// cmp.Equal short-circuits if it sees nil maps - but we don't want that...
```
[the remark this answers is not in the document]
metlos: We don't merely want to equate empty and nil maps but we also want to ignore certain keys in the non-nil maps while doing the comparison. [EquateEmpty](https://github.com/google/go-cmp/blob/master/cmp/cmpopts/equate.go#L23) merely equates empty and nil maps.

controllers/enqueue.go
metlos: I was hoping to extract some of the logic that we have in our watchers into a separate file in line with our aim to split up the reconcilers. At the same time, I didn't want to dilute this PR with a refactoring.

controllers/spiaccesstokenbinding_controller.go
```diff
@@ -54,23 +56,67 @@ import (
 
 	api "github.com/redhat-appstudio/service-provider-integration-operator/api/v1beta1"
 	"github.com/redhat-appstudio/service-provider-integration-operator/pkg/serviceprovider"
+	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
 )
 
-const linkedSecretsFinalizerName = "spi.appstudio.redhat.com/linked-secrets" //#nosec G101 -- false positive, this is not a private data
-
-var spiAccessTokenBindingLog = log.Log.WithName("spiaccesstokenbinding-controller")
+const deprecatedLinkedSecretsFinalizerName = "spi.appstudio.redhat.com/linked-secrets" //#nosec G101 -- false positive, this is not a private data
```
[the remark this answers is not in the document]
metlos: https://issues.redhat.com/browse/SVPI-349

sparkoo (approved):
Considering the circumstances, I think all my comments can be done as followups. I'm most worried about memory consumption with watching ServiceAccounts, so please make sure that there is enough memory cushion in the deployment.

xbaran4 (approved):

metlos:
I've updated the code to the new shape of the service account definition as agreed upon today in the chat.

Things still to do:
* update the docs to reflect the new shape and functionality
* more unit tests
* more integration tests
* actual manual testing on OpenShift and Kubernetes

sparkoo:
```
kind: SPIAccessTokenBinding
metadata:
spec:
  permissions:
    required:
      - type: rw
        area: registry
  repoUrl: https://quay.io/repository/mvala/spi-check
  secret:
    type: kubernetes.io/dockerconfigjson
    linkedTo:
    - serviceAccount:
        as: imagePullSecret
        managed:
          generateName: test-sa-
```

```
λ k apply -f samples/spiaccesstokenbinding-2.yaml 
Error from server (BadRequest): error when creating "samples/spiaccesstokenbinding-2.yaml": SPIAccessTokenBinding in version "v1beta1" cannot be handled as a SPIAccessTokenBinding: strict decoding error: unknown field "spec.secret.linkedTo[0].serviceAccount"
```

controllers/spiaccesstokenbinding_controller.go
metlos: I've looked at the graphs and the initial spike peaks at 112MB on an otherwise empty OpenShift cluster, while the latest operator image from main peaked at 94MB.

There are 351 service accounts in 69 namespaces in that cluster.

I guess we will have to wait and see how this pans out.

gazarenkov (approved):

PR #411, opened 2023-01-23
configurable Vault path prefix

sparkoo:
### What does this PR do?
Option to configure vault path prefix. Default is still `spi` so without explicitly setting it, everything should stay same.
 - new cli arg for both operator and oauth `--vault-data-path-prefix` env `VAULTDATAPATHPREFIX`
 - new configurable `hack/vault-init.sh` env variable `SPI_DATA_PATH_PREFIX`

https://issues.redhat.com/browse/SVPI-259

### How to test this PR?

images:
 - `quay.io/mvala/service-provider-integration-operator:svpi259vaultPrefix`
 - `quay.io/mvala/service-provider-integration-oauth:svpi259vaultPrefix`

Let's say you want to have data at `some/test/path`

1. deploy with images ^ with env variable `SPI_DATA_PATH_PREFIX` set to some path (for example on minikube something like `make deploy_minikube SPI_IMG_BASE=quay.io/mvala TAG_NAME=svpi259vaultPrefix SPI_DATA_PATH_PREFIX=some/test/path`)
2. set env variable in shared configmap for SPI pods `kubectl patch cm -n spi-system spi-shared-environment-config --type json --patch '[{"op": "add", "path": "/data/VAULTDATAPATHPREFIX", "value": "some/test/path"}]'`
3. try some action that writes data to Vault, like manual token upload or anything
4. to check you're really storing data at configured path, you can either connect to vault pod shell and use `vault` cli, or enable debug logs in SPI pods and see vault path in logs


docs/ADMIN.md
```diff
@@ -65,6 +65,7 @@ and oauth service to have the same value for them):
...
 | --vault-k8s-sa-token-filepath                         | VAULTKUBERNETESSATOKENFILEPATH |                       | Used with Vault kubernetes authentication. Filepath to kubernetes ServiceAccount token. When empty, Vault configuration uses default k8s path. No need to set when running in k8s deployment, useful mostly for local development. |
 | --vault-k8s-role                                      | VAULTKUBERNETESROLE            |                       | Used with Vault kubernetes authentication. Vault authentication role set for k8s ServiceAccount.                                                                                                                                   |
+| --vault-data-path-prefix                              | VAULTDATAPATHPREFIX            | spi                   | Path prefix in Vault token storage under which all SPI data will be stored. No leading or trailing '/' should be used, it will be trimmed.                                                                                         |
```
gazarenkov, line 68: maybe it is worth to mention (maybe not right here but somewhere in the doc) that  expected path would be like /{prefix}/data/ ?

mshaposhnik (approved):

docs/ADMIN.md
sparkoo: no, it's just prefix. the `data` part is handled in code https://github.com/redhat-appstudio/service-provider-integration-operator/pull/411/files#diff-9690267b7b50c62307b865c258b7a407dca24db90457879bc6c43a266a9546cfR39

xbaran4 (approved):

PR #418, opened 2023-01-25
feat: adding auth field in dockerconfigjson secret type

xbaran4:

### What does this PR do?
$TITLE

fixes: https://issues.redhat.com/browse/SVPI-308

### How to test this PR?

1. Deploy with SPIO_IMG=quay.io/pbaran/spio:SVPI-308-config
2. Setup quay oauth config
3. Apply `samples/binding-dockerconfigjson.yaml` but with your repo URL which has scheme (without scheme currently doesn't work) 
4. Go through oauth flow
5. Check the Injected secret. It should be of type dockerconfigjson like this:
image
The data should be similar to this:
```json
{
	"auths": {
		"quay.io": {
			"auth": "JG9hdXRodG9rZW46T2RqYWdVY3E3UVpqNEhkUUo5QXJCaURxxxxxxxxcg=="
		}
	}
}
```
6. create pod based on private image from quay.io and check that it was pulled successfully:
```yaml
apiVersion: v1
kind: Pod
metadata:
  name: private-reg
spec:
  containers:
  - name: private-reg-container
    image: quay.io/<your-private-image>
  imagePullSecrets:
  - name: test-secret-dockerconfigjson
``` 

#### Running a Tekton Task:

1. Go through the steps above which should result in `test-secret-dockerconfigjson` secret.
2. Deploy Tekton to minikube:
```shell
kubectl apply --filename https://storage.googleapis.com/tekton-releases/pipeline/latest/release.yaml
```
3. Create Scopeo task. This Tekton Task copies container images from and to a remote registry.
```shell
kubectl apply -f https://api.hub.tekton.dev/v1/resource/tekton/task/skopeo-copy/0.2/raw
````
4. Create a service account for the TaskRun referencing the docker config json secret:
```shell
apiVersion: v1
kind: ServiceAccount
metadata:
  name: tekton-task-service-account
secrets:
  - name: test-secret-dockerconfigjson
EOF
```
5. Create TaskRun:
```shell
apiVersion: tekton.dev/v1beta1
kind: TaskRun
metadata:
  name: skopeo-run
spec:
  serviceAccountName: tekton-task-service-account
  taskRef:
    name: skopeo-copy
  params:
  - name: srcImageURL
    value: docker://quay.io/testingorganization/che-server:latest # replace with your private src img
  - name: destImageURL
    value:  docker://quay.io/testingorganization/che-server:skopeo # replace with your dest img
  workspaces:
  - name: images-url
    emptyDir: {}
EOF
```
6. The TaskRun should be successful:
image


xbaran4:
> wdyt about trying to experiment with `~/github.com/docker/cli@v20.10.18+incompatible/cli/config/configfile/file.go` ConfigFile struct instead of working with strings? See
> 
> * `func (configFile *ConfigFile) SaveToWriter(writer io.Writer) error {`
> * `authCopy.Auth = encodeAuth(&authCopy)`

Something like this?

pkg/serviceprovider/access_token_mapper.go
```diff
@@ -54,23 +58,27 @@ func (at AccessTokenMapper) ToSecretType(secretType corev1.SecretType, mapping *
...
-			host = parsed.Host
+		if err != nil {
+			return nil, fmt.Errorf("failed to parse service provider url: %w", err)
+		}
+		host := parsed.Host
+		if host == "" {
+			host = at.ServiceProviderUrl
+		}
+		buff := new(bytes.Buffer)
+		conf := configfile.ConfigFile{AuthConfigs: map[string]types.AuthConfig{host: {Username: at.ServiceProviderUserName, Password: at.Token}}}
```
sparkoo, line 70: I don't like `docker/cli` dependency. I would be ok-ish with `moby/api/types/registry/AuthConfig` https://github.com/moby/moby/blob/master/api/types/registry/authconfig.go#L16 but it has no `auths` map.
I would propose this-like function:
```
func (at AccessTokenMapper) encodeDockerconfig() (string, error) {
	encoded := base64.StdEncoding.EncodeToString([]byte(at.ServiceProviderUserName + ":" + at.Token))
	host := ...

	type Auths map[string]struct {
		Auth string `json:"auth"`
	}
	eh := struct {
		Auths Auths `json:"auths"`
	}{
		Auths: Auths{
			host: {
				Auth: string(encoded),
			},
		},
	}

	jsonBytes, err := json.MarshalIndent(eh, "", "\t")
	if err != nil {
		return "", err
	}
	return string(jsonBytes), nil
}
```

pkg/serviceprovider/access_token_mapper.go
```diff
@@ -54,23 +58,27 @@ func (at AccessTokenMapper) ToSecretType(secretType corev1.SecretType, mapping *
...
-		host := at.ServiceProviderUrl
 		parsed, err := url.Parse(at.ServiceProviderUrl)
-		if err == nil && parsed.Host != "" {
-			host = parsed.Host
+		if err != nil {
+			return nil, fmt.Errorf("failed to parse service provider url: %w", err)
+		}
+		host := parsed.Host
+		if host == "" {
+			host = at.ServiceProviderUrl
```
sparkoo, line 67: I don't undestand this. In what cases it can happen and why we're writing `ServiceProviderUrl` to host?

pkg/serviceprovider/access_token_mapper.go
mshaposhnik: +1, even if it is puts kind of support on us :)

mshaposhnik:
Works in general. Pls consider @sparkoo fix of that dep-s.
image

mshaposhnik (approved):

scoheb:
I ran with my instance using the SPI image of  quay.io/pbaran/spio:SVPI-308-config and it works as expected!

FYI @skabashnyuk

pkg/serviceprovider/access_token_mapper.go
metlos: +1 for not using a dependency just because of a single struct that describes the format of a config file that is very unlikely to change.

metlos (approved):

sparkoo (approved):

PR #429, opened 2023-02-02
exclude test files from code coverage

sparkoo:
### What does this PR do?
$SUBJECT

n/a

### How to test this PR?

1. `make test`
2. check `cover.out` that it does not contain any test files

### DIFF
--- a/hack/coverage-sanitize.sh
+++ b/hack/coverage-sanitize.sh
@@ -12,7 +12,8 @@ if [ ! -f ${COVER_FILE} ]; then
 fi
 
 echo "creating backup of ${COVER_FILE} to ${COVER_FILE}_bck"
-cp ${COVER_FILE} ${COVER_FILE}_bck
+mkdir -p ${REPO_DIR}/.tmp
+cp ${COVER_FILE} ${REPO_DIR}/.tmp/cover.out_bck
 
 declare -a EXCLUDE=("integration_tests" "test_util.go" "testtokenstorage.go")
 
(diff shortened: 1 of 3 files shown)


hack/coverage-sanitize.sh
```diff
@@ -0,0 +1,22 @@
...
+
+
+THIS_DIR="$( dirname "$(realpath "$0")")"
+REPO_DIR="$( realpath ${THIS_DIR}/../ )"
+
+COVER_FILE=${REPO_DIR}/cover.out
+
+if [ ! -f ${COVER_FILE} ]; then
+  echo "${COVER_FILE} not found, nothing to do"
+  exit 1
```
metlos, line 11: is this an error that should fail the build?

sparkoo: good question. I would say yes, because it's run mainly from makefile immediately after `go test` which is supposed to create `cover.out`. If it doesn't do that, something went really wrong.
However, I don't have strong opinion about this. If you have good reason not to do that, I'll change it to `0` or handle it in the makefile to not fail in case of non-zero result.

mshaposhnik (approved):

sparkoo:
> what do you think(have you tried) https://docs.codecov.com/docs/codecovyml-reference#ignore?

I have not found this. It looks better than my script.

sparkoo:
closing in favor of https://github.com/redhat-appstudio/service-provider-integration-operator/pull/430

PR #433, opened 2023-02-06
move storage implementations to own packages

sparkoo:
### What does this PR do?
Move vault token storage into `vaultstorage` package and secret storage into `secretstorage` package. This is to prepare to add new aws storage and keep it more separate and clean.

prepare to https://issues.redhat.com/browse/SVPI-342

### How to test this PR?

```
quay.io/redhat-appstudio/pull-request-builds:spi-operator-433
quay.io/redhat-appstudio/pull-request-builds:spi-oauth-433
```

Just refactoring, evertyhing must work as before

### DIFF
--- a/cmd/oauth/oauth.go
+++ b/cmd/oauth/oauth.go
@@ -38,6 +38,7 @@ import (
 	oauth2 "github.com/redhat-appstudio/service-provider-integration-operator/pkg/serviceprovider/oauth"
 	"github.com/redhat-appstudio/service-provider-integration-operator/pkg/spi-shared/config"
 	"github.com/redhat-appstudio/service-provider-integration-operator/pkg/spi-shared/tokenstorage"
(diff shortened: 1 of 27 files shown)


metlos (commented):
I agree with the idea, but it opens up 1 code smell that we have in the code base and that is the use of Vault-based TestTokenStorage across the testsuite... IMHO a "test token storage" should abstract away the backend, not expose it.

But that would probably be better done as a separate PR - would pls create an issue for that?

sparkoo:
> I agree with the idea, but it highlights 1 code smell that we have in the code base and that is the use of Vault-based TestTokenStorage across the testsuite... IMHO a "test token storage" should abstract away the backend, not expose it.
> 
> But that would probably be better done as a separate PR - would pls create an issue for that?

https://issues.redhat.com/browse/SVPI-350

xbaran4 (approved):

PR #441, opened 2023-02-08
Prepare CLI for multiple tokenstorage implementations

sparkoo:
### What does this PR do?

Refactor CLI args and CLI related code:
 - shared cli args and common functions moved out to `cmd/` (BASE)
 - operator cli args and common functions moved out to `cmd/operator/cli` (SERVICE)
 - oauth cliargs and common functions moved out to `cmd/oauth/cli` (SERVICE)
 - vault specific cli args and functions moved out to `pkg/spi-shared/tokenstorage/vaultstorage/vaultcli` (COMPONENT)

logic/rules:
 - COMPONENT cli must not have dependency on any other cli
 - BASE cli have dependency on common COMPONENTs cli
 - SERVICE cli have dependency on BASE cli

This PR also adds new cli/env property `--tokenstorage` (`TOKENSTORAGE`), with now only and default option `vault`. Code is prepared for other implementations.

prepare https://issues.redhat.com/browse/SVPI-342

### How to test this PR?

```
quay.io/redhat-appstudio/pull-request-builds:spi-operator-441
quay.io/redhat-appstudio/pull-request-builds:spi-oauth-441
```

All CLI must work the same. If `--tokenstorage` set to something else than `vault`, it must fail to start with proper message.


oauth.Dockerfile
```diff
@@ -9,13 +9,14 @@ COPY go.sum go.sum
...
 
 # Copy the go source
-COPY cmd/oauth cmd/oauth
+COPY cmd/ cmd/
 COPY api/ api/
-COPY oauth/ oauth/
 COPY pkg/ pkg/
+COPY oauth/ oauth/
+
+COPY static/ static/
```
xbaran4, line 19: Isn't it better to keep this step before copying the source code? To utilize caching as the static folder is rarely (never) changed.

xbaran4 (approved):

oauth.Dockerfile
sparkoo: yes, but cache is shared with operator dockerfile until line 17 and moving this up would break it and I see more benefit in shared cache with operator. 
However, I think there is place for improvement but tbh I'm not sure it's worth it. We can: 
 - move it just above `COPY oauth/ oauth/`, but I think it has very little benefit
 - move it in front of sources and do the same in operator as well, but we don't need it to build the operator
 - keep it like this patch proposes

WDYT?

xbaran4: > yes, but cache is shared with operator dockerfile until line 17 and moving this up would break it and I see more benefit in shared cache with operator. 
> 
Ah yes good thinking.

> * move it just above `COPY oauth/ oauth/`, but I think it has very little benefit

I think I like this the most.

> * move it in front of sources and do the same in operator as well, but we don't need it to build the operator

This would probably be the most efficient but I don't like having the statics in the operator when we don't need them.

As you pointed out it's not much of a difference so feel free to pick what you find best.

sparkoo: ok, I've move static above copy oauth

PR #442, opened 2023-02-09
Ensure that binding sample without scheme is valid.

xbaran4:
### What does this PR do?
When repoUrl does not contain a scheme we cannot properly parse the host part of it. We need to do this, for example, to match the OAuth configuration or create Host Credentials SP.
Instead of checking this throughout the code base, it can be checked asap in one place.

This PR checks the repoUrl at the beginning of reconciliation and tries to prepend https scheme if the host cannot be parsed. If the host still cannot be parsed then updates the binding with an error.

Note that this PR allows the user to create any binding without the scheme. Then https scheme is assumed and the binding will be valid. If only Quay bindings should be allowed without scheme then the PR can be updated. However, in the future when Quay may have a non-default URL (not quay.io) this may be hard to assure.

https://issues.redhat.com/browse/SVPI-306

### How to test this PR?

1. `kubectl apply -f samples/binding-dockerconfigjson.yaml` should not result in an error phase and the repoUrl should change to contain `https` scheme.
2. Add `https://` to the URL of the sample above.  and apply it. It should behave exactly the same as without a scheme.
3. Change the whole URL of the sample above to `:://abc` it should end up in an error phase.

Operator image built by me: 
```
quay.io/pbaran/spio:SVPI-306-sample
```

```
quay.io/redhat-appstudio/pull-request-builds:spi-operator-#
```


controllers/spiaccesstokenbinding_controller.go
```diff
@@ -589,6 +594,52 @@ func (r *SPIAccessTokenBindingReconciler) updateBindingStatusSuccess(ctx context
 	return nil
 }
 
+// assureProperValuesInBinding updates the binding, replacing values which we are valid in multiple formats, but we would
+// like to work with some specific format in the rest of the codebase.
+// For example, we allow the repoUrl to be with or without scheme, but we need to parse out the host and for that we need the url to contain scheme.
```
sparkoo, line 599: can you please comment what return value means?

sparkoo (commented):
wdyt about fail in case scheme is not `https` or empty? Or maybe better to fail if it is `http`?

sparkoo (approved):

xbaran4:
> wdyt about fail in case scheme is not `https` or empty? Or maybe better to fail if it is `http`?

I've created a new issue to invalidate binding with `http` https://issues.redhat.com/projects/SVPI/issues/SVPI-352

metlos (approved):

PR #445, opened 2023-02-09
Compensate the changes in the dependent objects in case of the binding update failure

metlos:
### What does this PR do?

This deletes the dependent objects (secret(s) and service account(s)) created for a binding if there is a failure during binding reconciliation after those objects have been created.

This makes sure that we don't create "garbage" until all the operations "click" in the right order.

In addition to recovering from binding update failure, this also tries to solve the issue that was especially apparent on OpenShift, which automatically injects an image pull secret to any service account. This immediate update to a newly created SA wasn't accounted for in our code and so this PR adds a retry loop that tries to reload the SA from the cluster and retry the linking (which happens after a create of the SA).

https://issues.redhat.com/browse/SVPI-210
https://issues.redhat.com/browse/SVPI-351

### How to test this PR?
This is especially reproducible on OpenShift with service accounts.

1. Create a binding with a managed service account definition and secret without name so that generated names are used.
2. Provide the data for the token either using OAuth flow or manual upload so that the binding moves to the injected state.
3. observe that there is only 1 new secret and 1 new service account as per definition

```
quay.io/redhat-appstudio/pull-request-builds:spi-operator-445
quay.io/redhat-appstudio/pull-request-builds:spi-oauth-445
```


controllers/bindings/dependents.go
```diff
@@ -32,20 +33,90 @@ type DependentsHandler struct {
...
+// It can be used in the DependentsHandler.RevertTo method to delete the secret/service accounts from the cluster
+// that have been created after an instance of this struct has been returned from the DependentsHandler.CheckPoint
+// method.
+type CheckPoint struct {
+	secretName          string
+	serviceAccountNames serviceAccountNamesAndLinkTypes
+}
+
+func (s serviceAccountNamesAndLinkTypes) containsName(name string) bool {
+	for k := range s {
```
sparkoo, line 59: I don't understand this. Is it checking that map contains some key? How is it different from 
```
if serviceAccount, contains := checkPoint.serviceAccountNames[sa.Name]; contains {
  ...
  if serviceAccount.secret {
    ...
  }
  ...
} else {
  ...
}
```

metlos: Ah, yes, good catch.

sparkoo (approved):
have you seen this "transaction" approach anywhere as a good practice to solve our issue in kubernetes operators? To me it feels like we're only trying to patch some bigger design issue, but not actually solving it. However, I don't have better solution now.

mshaposhnik (approved):

metlos:
> have you seen this "transaction" approach anywhere as a good practice to solve our issue in kubernetes operators? To me it feels like we're only trying to patch some bigger design issue, but not actually solving it. However, I don't have better solution now.

I have exactly the same feeling. One approach might be the one suggested by @skabashnyuk - to exit the reconcile after any update to the cluster and continue from that point on. But that approach will deal with the very same problems that this "transaction" approach is dealing with, too. It would definitely be more clean solution in Kubernetes, but would require much more drastic refactoring that I didn't want to introduce right now.

PR #449, opened 2023-02-10
Https URLs validate

mshaposhnik:
### What does this PR do?
Adds validator which making sure we're always using `https` whenever this possible in the manager, storage, static or dynamic service provider configurations.

https://issues.redhat.com/browse/SVPI-344

### How to test this PR?

```
quay.io/redhat-appstudio/pull-request-builds:spi-operator-#
```
build own or use `mshaposh/spio:latest` + `mshaposh/spis:latest`

Static configurations  (all actions are performed in **spi-system** namespace) :

 - Change either `BASEURL` or `VAULTHOST` to `http://.....` in `spi-shared-environment-config`  map must lead to both manager and oauth are failed to start:
 `
"error":"failed to load the configuration file from /etc/spi/config.yaml: service configuration validation failed: struct validation failed: Key: 'SharedConfiguration.BaseUrl' Error:Field validation for 'BaseUrl' failed on the 'https_only' tag"
 `
 
 - Update `shared-configuration-file` secret with the following dummy **http** type provider:
 ```
 - type: GitLab
  clientId: foo
  clientSecret: bar
  baseUrl: http://dummy-gitlab.com
  ```
  Similar error must be on startup of both pods:
  `
  service configuration validation failed: struct validation failed: Key: 'SharedConfiguration.ServiceProviders[2].ServiceProviderBaseUrl' Error:Field validation for 'ServiceProviderBaseUrl' failed on the 'https_only' tag
  `
 
 
Runtime configurations (all actions are performed in **user** namespace) : 

- Create secret as follows:
```
kind: Secret
apiVersion: v1
metadata:
  name: my-gitlab
  labels:
    spi.appstudio.redhat.com/service-provider-type: GitLab
    spi.appstudio.redhat.com/service-provider-host: dummy-gitlab.com
data:
  clientId: <base64 foo >
  clientSecret: <base64 bar >
  authUrl: <base64 http://dummy-gitlab.com/auth/>
  tokenUrl: <base64 http://dummy-gitlab.com/token/>
```

- Create any kind of SPI CR for given URL  (binding, access check, file request....) with that host configured over HTTP:

```
kind: SPIAccessTokenBinding
metadata:
  namespace: default
spec:
  permissions:
    required:
      - type: rw
        area: registry
  repoUrl: http://dummy-gitlab.com
  secret:
```

the CR must be set into error state:

```
errorMessage: >-
    failed to create runtime configuration for service provider GitLab: struct validation
    failed: Key: 'ServiceProviderConfiguration.ServiceProviderBaseUrl'
    Error:Field validation for 'ServiceProviderBaseUrl' failed on the
    'https_only' tag
  errorReason: UnsupportedServiceProviderConfiguration
  linkedAccessTokenName: ''
  phase: Error
```

At the same time, similar CR congirured with HTTPS must work fine.


sparkoo:
+1 from me to use annotation based validation on file configs and cli arguments. Basically to validate critical things that we must pass to start or fail asap with clear message. We already have there annotations to e.g. default value, so I think it make sense to think about validation in the same context.
However, I'm not sure how this is useful for validation of live objects, like our CRs, user SP config secrets or other configuration secrets (token storage credentials, sp configuration). I would be very careful there.

mshaposhnik:
I do know that error messages are far from ideal. There is ability in the validator library to register custom error translations. but it is not looks very simple,so may require some additional efforts of coding/debugging. We may do that as a follow-up.

controllers/spiaccesstoken_controller.go
```diff
@@ -195,7 +197,14 @@ func (r *SPIAccessTokenReconciler) Reconcile(ctx context.Context, req ctrl.Reque
 	// persist the SP-specific state so that it is available as soon as the token flips to the ready state.
 	sp, err := r.ServiceProviderFactory.FromRepoUrl(ctx, at.Spec.ServiceProviderUrl, req.Namespace)
 	if err != nil {
-		if uerr := r.flipToExceptionalPhase(ctx, &at, api.SPIAccessTokenPhaseError, api.SPIAccessTokenErrorReasonUnknownServiceProvider, err); uerr != nil {
+		var reason api.SPIAccessTokenErrorReason
+		var validationErr validator.ValidationErrors
+		if stderrors.As(err, &validationErr) {
+			reason = api.SPIAccessTokenErrorUnsupportedServiceProviderConfiguration
+		} else {
```
xbaran4, line 204: Is it necessary here (and in other places) to have the new ErrorReason? I am not opposed, but the number of different reasons keeps rising and here it makes the code more "complex", and I think it might be enough to just keep it as `SPIAccessTokenErrorReasonUnknownServiceProvider` and explain more context in the ErrorMessage. WDYT?

xbaran4 (approved):
LGTM. Just wanted to share my view on the rising number of ErrorReasons.

metlos (approved):

sparkoo (approved):

controllers/spifilecontentrequest_controller.go
```diff
@@ -188,10 +191,16 @@ func (r *SPIFileContentRequestReconciler) Reconcile(ctx context.Context, req ctr
 		// check if the URL is processable and create binding, otherwise fail fast
 		sp, err := r.ServiceProviderFactory.FromRepoUrl(ctx, request.Spec.RepoUrl, request.Namespace)
 		if err != nil {
-			lg.Error(err, "unable to get the service provider for the SPIFileContentRequest")
-			// we determine the service provider from the URL in the spec. If we can't do that, nothing works until the
-			// user fixes that URL. So no need to repeat the reconciliation and therefore no error returned here.
-			r.updateFileRequestStatusError(ctx, &request, noSuitableServiceProviderFound)
+			var validationErr validator.ValidationErrors
+			if stderrors.As(err, &validationErr) {
+				lg.Error(err, "unable to validate the service provider for the SPIFileContentRequest")
```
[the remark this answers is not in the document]
mshaposhnik: it should unwrap the  error., so we'll see the details. I don't want to be so specific there because potentially other kind of validations may be done as well

controllers/spiaccesstoken_controller.go
mshaposhnik: Good question. I understand your point. At the same time, it we will need to find CR-s to fix or count them etc. 
separate status is better imo.

PR #467, opened 2023-02-21
ITest improvements

metlos:
### What does this PR do?
* Make it possible to log the interactions with K8s client
* Improve configurability of Capabilities in integration tests
* move service provider mock out of integration tests so that it is usable elsewhere, too.

none

### How to test this PR?
test suite should pass


controllers/logging_client.go
```diff
@@ -0,0 +1,313 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//go:build !release
+
+package controllers
```
sparkoo, line 17: isn't `integrationtests` better place for this file? or some other test package?

metlos: My original intention was to also use this outside of integration tests but since currently it is only being used there and frankly there is little incentive to use it in normal unit tests that have much tighter control over the K8s calls, I think you're right. I moved it to `integrationtests` package.

PR #477, opened 2023-02-24
make tokenuploader create new SPIAccessToken

gazarenkov:
### What does this PR do?
Creates new SPIAccessToken, associates it with uploaded with Secret one and moves SPIAccessToken to Ready state.

n/a

Makes it possible to not only associate token with existed SPIaccessToken but create and make them ready to use right away. It simplifies the flow for Admin/DevOPs who need to create simple secret (for example by initalization script) but do not use it, so not needs SPIAccessTokenBinding. 

https://issues.redhat.com/browse/SVPI-382

### How to test this PR?
There is dedicated integration test included, to run it call `make itest focus="TokenUploadController"`

To reproduce it manually
- make sure SPIOperator configured with `ENABLETOKENUPLOAD=true` (see Admin Guide for details)
- Obtain Access Token ($AT) from the Service Provider of your choice. For example from GitHub->Settings->Developer Settings->Personal Access Tokens  
- Create Secret l: `kubectl apply -f my.yaml` 
With `my.yaml `file content like: 
` 
```
apiVersion: v1
kind: Secret
metadata:
  name: upload-secret
  labels:
    spi.appstudio.redhat.com/upload-secret: token
    spi.appstudio.redhat.com/token-name: my-spi-access-token
type: Opaque
stringData:
  providerUrl: https://github.com
  tokenData: $AT
```
`
- Make sure SPIAccessToken is created and Ready state (may take some time): `kubectl get spiaccesstoken my-spi-access-token -o yaml `
(status:phase:Ready)


gazarenkov:
> 

There was an info about it in the docs but you are right, it needs to be more explained because it is not trivial since we want to make Secret deleted any way by security reason, so added "Error handling" chapter with extended information.

controllers/tokenupload_controller.go
```diff
@@ -61,67 +67,34 @@ type TokenUploadReconciler struct {
...
-		// if spiTokenName field is not empty - try to find SPIAccessToken by it
-		if uploadSecret.Labels[spiTokenNameLabel] != "" {
-			spiTokenName := uploadSecret.Labels[spiTokenNameLabel]
-			accessToken = spi.SPIAccessToken{}
-			err = r.Get(ctx, types.NamespacedName{Name: spiTokenName, Namespace: uploadSecret.Namespace}, &accessToken)
-
+		// try to find  SPIAccessToken
+		accessToken, err := findSpiAccessToken(ctx, uploadSecret, r, lg)
+		if err != nil {
+			continue
```
metlos, line 91: We usually accompany the error handling with logging the error. Here, it looks like we're not doing that when in fact we are because the `findSpiAccessToken` does that. Maybe worth a comment here that it has already been logged and why it is ok to not react in any other way to it...

controllers/tokenupload_controller.go
```diff
@@ -130,7 +103,8 @@ func (r *TokenUploadReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 		}
 
 		logs.AuditLog(ctx).Info("manual token upload initiated", uploadSecret.Namespace, accessToken.Name)
-		err = r.TokenStorage.Store(ctx, &accessToken, &token)
+		// Upload Token, it will cause update SPIAccessToken State as well
+		err = r.TokenStorage.Store(ctx, accessToken, &token)
 		if err != nil {
 			logError(ctx, uploadSecret, fmt.Errorf("token storing failed: %w", err), r, lg)
 			logs.AuditLog(ctx).Error(err, "manual token upload failed", uploadSecret.Namespace, accessToken.Name)
```
metlos, line 110: This audit log message is IMHO confusing. This is not a manual token upload per se, but rather reading the token data from an annotated secret. IMHO we should distinguish these two cases.

We should make this audit log, because it happened during the access to the 3rd party system (the token storage), but I think we should be able to distinguish between a manual token upload using the REST API and using a Secret in the logs.

controllers/tokenupload_controller.go
```diff
@@ -217,35 +191,57 @@ func tryDeleteEvent(ctx context.Context, secretName string, ns string, r *TokenU
...
-func findTokenByUrl(ctx context.Context, url string, ns string, r *TokenUploadReconciler, lg logr.Logger) *spi.SPIAccessToken {
+func findSpiAccessToken(ctx context.Context, uploadSecret corev1.Secret, r *TokenUploadReconciler, lg logr.Logger) (*spi.SPIAccessToken, error) {
+	spiTokenName := uploadSecret.Labels[spiTokenNameLabel]
 
-	tokenList := spi.SPIAccessTokenList{}
-	if err := r.List(ctx, &tokenList, client.InNamespace(ns)); err != nil {
-		lg.Error(err, "Can not get tokenList of tokens ")
-		return nil
+	if spiTokenName == "" {
+		lg.V(logs.DebugLevel).Info("No label found, will try to create with generated ", spiTokenNameLabel, spiTokenName)
```
metlos, line 198: This looks like an unfinished info message? Also, this seems to inform about a logic that is then being done in another method, so maybe we should postpone the logging and do it at a place that is closer to the actual logic being applied so that there's a better chance of the two being in sync in the future :)

controllers/tokenupload_controller.go
```diff
@@ -204,6 +176,8 @@ func logError(ctx context.Context, secret corev1.Secret, err error, r *TokenUplo
 
 }
 
+// Contract: having at only one event if upload failed and no events if uploaded.
+// For this need to delete the event every attempt
 func tryDeleteEvent(ctx context.Context, secretName string, ns string, r *TokenUploadReconciler, lg logr.Logger) {
```
metlos, line 181: Why passing the reconciler as a parameter? Could this be a member method of the reconciler instead?

controllers/tokenupload_controller.go
```diff
@@ -130,7 +103,8 @@ func (r *TokenUploadReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 		}
 
 		logs.AuditLog(ctx).Info("manual token upload initiated", uploadSecret.Namespace, accessToken.Name)
-		err = r.TokenStorage.Store(ctx, &accessToken, &token)
+		// Upload Token, it will cause update SPIAccessToken State as well
+		err = r.TokenStorage.Store(ctx, accessToken, &token)
 		if err != nil {
 			logError(ctx, uploadSecret, fmt.Errorf("token storing failed: %w", err), r, lg)
```
metlos, line 109: "failed to store the token: %w" would IMHO sound more natural...

docs/USER.md
```diff
@@ -141,25 +141,31 @@ There is an ability to upload Personal Access Token using very short living K8s
...
-- Create Kubernetes Secret with labeled with `label.spi.appstudio.redhat.com/upload-secret: token` and `spi.appstudio.redhat.com/token-name: $TOKEN_NAME`:
+- Create Kubernetes Secret with labeled with `spi.appstudio.redhat.com/upload-secret: token` and `spi.appstudio.redhat.com/token-name: $TOKEN_NAME`:
+- If you want new SPIAccessToken to be created and associate it with the Token, so the SPIAccessToken will be Ready right away, put the name of (non-existed) SPIAccessToken you want to be assigned to the `spi.appstudio.redhat.com/token-name` label or just remove this label (or leave it empty), in the latter case SPIAccessToken name will be randomly generated. 
 
 ```yaml
 apiVersion: v1
 kind: Secret
 metadata:
-  name: upload-secret
+  name: $upload-secret-name
```
metlos, line 154: nitpick: variables are usually upper case to distinguish them more clearly from the rest of the text.

controllers/tokenupload_controller.go
gazarenkov: @metlos 
Agree, but I would say, both plain REST and K8s API (Secret) REST ways are manual (in contrast to OAuth way), no?
So, we probably need to avoid "manual" at all or have something "manual REST" vs "manual K8s"?
WDYT?

controllers/tokenupload_controller.go
gazarenkov: I see, if it is an agreement, I can move logging here

controllers/tokenupload_controller.go
gazarenkov: OK for unfnished.
The second one not sure I understand... Do you mean it is better to have another check like if result==nil && err==nil after the function call just for putting debug log message?

controllers/tokenupload_controller.go
gazarenkov: sure :)

gazarenkov:
> @gazarenkov I was able to get an error explanation for the case of missing providerUrl. But I'd failed to do that for the case of incorrect URL like
> 
> ```
> cat <<EOF | kubectl apply -n $TARGET_NAMESPACE -f -
> apiVersion: v1
> kind: Secret
> metadata:
>   name: upload-secret
>   labels:
>     spi.appstudio.redhat.com/upload-secret: token
>     spi.appstudio.redhat.com/token-name: my-spi-access-token
> type: Opaque
> stringData:
>   providerUrl: lasdjkfoiasodfij
>   tokenData: ${GITHUB_SPI}
> EOF
> ```
> 
> or invalid token name
> 
> ```
> cat <<EOF | kubectl apply -n $TARGET_NAMESPACE -f -
> apiVersion: v1
> kind: Secret
> metadata:
>   name: upload-secret
>   labels:
>     spi.appstudio.redhat.com/upload-secret: token
>     spi.appstudio.redhat.com/token-name: my-spi-access-token_2023_03_02__15_37_28
> type: Opaque
> stringData:
>   providerUrl: https://github.com
>   tokenData: ${GITHUB_SPI}
> EOF
> ```
> 
> Also, what do you think about checking that providerUrl is a valid url with https protocol?

@skabashnyuk it was a bug with invalid token name, it validated by K8s, but did not correctly processed, thanks for pointing it out, fixed.
As for providerUrl, I think it's validity should be checked on CR type / spiaccesscontroller level validation, not here. Removed check for its emptiness for consistency.

metlos (approved):

PR #482, opened 2023-03-01
Svpi366 aws deploy

sparkoo:
### What does this PR do?
Shuffle kustomize yamls to have now 4 options of deployment:
 - minikube_vault
 - minikube_aws
 - openshift_vault
 - openshift_aws

I couldn't find a way how to do this better with kustomize. This approach is prone to get out of hands if we will need more configuration options as it defines overlays for all permutations. Any ideas are welcome, but I think this is the limit of kustomize and we would need something like helm to make this more generically configurable. It now looks like this:
```
            ┌────────┐  ┌─────┐   ┌────┐    ┌───┐  ┌───┐
            │ manager│  │oauth│   │rbac│    │svc│  │...│
            └────▲───┘  └──▲──┘   └──▲─┘    └─▲─┘  └─▲─┘
                 │         │         │        │      │
                 └─────────┼─────────┼────────┴──────┘
                           │ default │
                           └───▲──▲──┘
                               │  │
                  ┌─────┐      │  │   ┌───────┐
      ┌──────────►│ aws ├──────┘  └───┤ vault │◄──────┐
      │           └──▲──┘             └──▲────┘       │                        ▲
      │              │                   │            │               bases    │
      │              │                   │            │            ────────────┤
      │              │                   │            │             overlays   │
┌─────┴──────┐ ┌─────┴───────┐  ┌────────┴─────┐ ┌────┴──────────┐             ▼
│minikube_aws│ │openshift_aws│  │minikube_vault│ │openshift_vault│
└────────────┘ └─────────────┘  └──────────────┘ └───────────────┘
```

Like this, in infra-deployments we will need to change from using `config/openshift` to `config/overlays/openshift_vault` and then with AWS to `config/overlays/openshift_aws`.

TODO:
 - `shared-configuration-file` is now prefixed as `spi-shared-configuration-file`, I need to fix this to avoid compatibility issues.
 - update makefile
 - update docs
 - test and prepare update for infra-deployments

https://issues.redhat.com/browse/SVPI-366

### How to test this PR?

I can imagine 2 tests:
 - `kustomize build` with new and old overlays and compare diff
   - old `openshift` should match new `overlays/openshift_vault`
   - old `minikube` should match new `overlays/minikube_vault`

 - deploy with any overlay at `config/overlays`, should work
   - aws deployments require `aws-secretsmanager-credentials` secret with `config` and `credentials` keys with values from `~/.aws/config` and `~/.aws/credentials`
     - like this: `kubectl create secret generic aws-secretsmanager-credentials --from-file=/Users/mvala/.aws/config --from-file=/Users/mvala/.aws/credentials -n spi-system`


sparkoo:
closing in favor of https://github.com/redhat-appstudio/service-provider-integration-operator/pull/502

PR #510, opened 2023-03-13
Svpi 389 github error rate metrics

xbaran4:
### What does this PR do?
Adds 2 counter metrics: 
1. counts occurrences of unexpected status codes when making an unauthorized request to get a GitHub repository. The counters are labeled depending on the status code
2. counts the number of rateLimitErrors occurred when making an authorized request to get a GitHub repository

image
image

https://issues.redhat.com/browse/SVPI-389

### How to test this PR?

1. create a personal access token on github and upload it to SPI (either through REST or through secret)
2. run the script bellow. The script prints your rate limit, then does 1000 request and then prints your rate limit again.
You might want to run 5 scripts in parallel to speed up the depletion of rate limit.
```shell
#!/bin/bash

curl 'https://api.github.com/rate_limit' -H "Authorization: bearer $YOUR_TOKEN"
for _ in {1..1000} ; do
    REPO=$(tr -dc 'a-z' < /dev/urandom | head -c 15) # I am creating random repo names here to avoid any caching
    USR=$(tr -dc 'a-z' < /dev/urandom | head -c 15)
    curl "https://api.github.com/$USR/$REPO" -H "Authorization: bearer $YOUR_TOKEN" >> /dev/null
#    sleep 1
done
curl 'https://api.github.com/rate_limit' -H "Authorization: bearer $YOUR_TOKEN"
```
3. after you exceeded rate limit, create access check sample with some private repository, it should end up with error seen in screenshot
`kubectly apply -f samples/spiaccesscheck.yaml`
4. ssh into operator container and run
```shell
curl localhost:8080/metrics | grep rate
````
you should see the rate limit metrics with value higher than 0 like on screenshot

```
```

### CODE REVIEW

**Review by metlos — APPROVED** [review]

File: pkg/serviceprovider/github/github.go [hunk 1]
```diff
@@ -225,9 +254,13 @@ func (g *Github) CheckRepositoryAccess(ctx context.Context, cl client.Client, ac
 		}
 		ghRepository, _, err := ghClient.Repositories.Get(ctx, owner, repo)
 		if err != nil {
+			var rateLimitError *github.RateLimitError
+			if errors.As(err, &rateLimitError) {
```
sparkoo, line 258: why not `errors.Is` ?

xbaran4:
> I think that is not all
> 
> 1. I've created PAT
> 2. I've used it to query github API and reached the limit
> 
> ```
>   "resources": {
>     "core": {
>       "limit": 5000,
>       "used": 5000,
>       "remaining": 0,
>       "reset": 1678876130
>     },
> ```
> 
> 3. I've created SPiAccesstToken
> 4. I've used rest to upload the token
> 5. Get the error
> 
> Знімок екрана 2023-03-15 о 11 58 55
> 
> 6. no metrics
> 
> ```
> ❯ kubectl -n spi-system exec deployment/spi-oauth-service -c oauth -t -- curl --silent localhost:8080/metrics | grep rate
> <nothing here>
> ```

the metrics are on operator container

mshaposhnik (approved):

pkg/serviceprovider/github/github.go
xbaran4: We just want to find out that its type is rateLimitError no?
> As(err error, target interface{}) bool checks if any error in the chain of wrapped errors matches the target. The difference is that this function checks whether the error has a specific type, unlike the Is(), which examines if it is a particular error object.

sparkoo: I see, thanks

sparkoo (approved):

xbaran4:
I've updated the PR so that every call using authorized GH client can update the rate limit error metric

PR #512, opened 2023-03-13
Flip token binding to awaiting state if data suddenly disappears & clean sync objects

mshaposhnik:
### What does this PR do?
Flips token binding to awaiting state if token data suddenly disappears from storage & clean synced objects (secrets & sa)

### What issues does this PR fix or reference?

https://issues.redhat.com/browse/SVPI-414

### How to test this PR?

 
-  Create binding, upload token.
-  Delete  token data from storage
-  Binding must appear in the `AwaitingTokenData` state

```
quay.io/redhat-appstudio/pull-request-builds:spi-controller-#
quay.io/redhat-appstudio/pull-request-builds:spi-oauth-#
```

PR #514, opened 2023-03-14
cleanup the token if it does not have any data

sparkoo:
### What does this PR do?
If token data disappears from token storage for some reason, this PR ensures that both `SPIAccessToken` and `SPIAccesstokenBinding` are flipped back to `AwaitingTokenData` state. The injected Secret itself is also removed.

https://issues.redhat.com/browse/SVPI-413
https://issues.redhat.com/browse/SVPI-414

related PR: https://github.com/redhat-appstudio/service-provider-integration-operator/pull/512

### How to test this PR?

```
quay.io/redhat-appstudio/pull-request-builds:spi-controller-514
quay.io/redhat-appstudio/pull-request-builds:spi-oauth-514
```

1. create SPIAccessTokenBinding
2. upload the token
3. delete data from token storage (e.g. connect to vault pod terminal and delete secret with `vault kv delete <path>`)
4. restart operator pod
5. check that `SPIAccessTokenBinding` and `SPIAccessToken` are in `AwaitingTokenData` and linked Secret is gone


mshaposhnik:
In general looks ok. I think we should test it separately from the case when _metadata_ disappears.
Also, pls fix the access token example, as it seems to be outdated.

controllers/spiaccesstoken_controller.go
```diff
@@ -151,7 +151,7 @@ func (r *SPIAccessTokenReconciler) Reconcile(ctx context.Context, req ctrl.Reque
 		return ctrl.Result{}, fmt.Errorf("failed to load the token from the cluster: %w", err)
 	}
 
-	lg = lg.WithValues("phase_at_reconcile_start", at.Status.Phase)
+	lg = lg.WithValues("phase_at_reconcile_start", at.Status.Phase, "accesstoken", at)
```
metlos, line 154: this puts the whole object into the log values. The log already has the object key set by the controller manager btw. IMHO it would be better to just add the data that is helping you here instead of dumping whole object into the log messages...

controllers/spiaccesstoken_controller.go
```diff
@@ -272,6 +276,18 @@ func (r *SPIAccessTokenReconciler) Reconcile(ctx context.Context, req ctrl.Reque
 	return ctrl.Result{RequeueAfter: r.durationUntilNextReconcile(&at)}, nil
 }
 
+func (r *SPIAccessTokenReconciler) reconcileTokenData(ctx context.Context, at *api.SPIAccessToken) error {
+	if at.Status.TokenMetadata != nil && at.Status.Phase == api.SPIAccessTokenPhaseReady {
```
metlos, line 280: Can we do this somehow without depending on the status? Or have a thorough explanation why doing it this way is ok?

Maybe it is, but I'm fearing we may be making the reconciliation not idempotent by these status checks.

sparkoo: technically it does not depend on state. This is just an optimization to not check the data for all tokens, but only for those which we expect to have some data (ready phase + metadata!=nil). I can remove this condition and it will work the same, but we will touch tokenstorage everytime, even in cases we expect no data there.

I'm not sure which option is more correct.

mshaposhnik (approved):

controllers/spiaccesstoken_controller.go
metlos: If the behavior is the same without this check and it is only an optimization then that's great. But if that's the case, I would prefer an explanation of the fact here, because it is not obvious at a glance (at least for me).

sparkoo: added comments

metlos (approved):

PR #517, opened 2023-03-14
Introduce the SecretStorage interface that generalizes the token storage to store arbitrary byte arrays.

metlos:
### What does this PR do?
Introduce the SecretStorage interface that generalizes the token storage to store arbitrary byte arrays.

The existing implementations of TokenStorage interface were converted to simple DefaultTokenStorage instances set up with appropriate SecretStorage implementations.

https://issues.redhat.com/browse/SVPI-371

### How to test this PR?
There should be no noticeable difference in functionality.

```
quay.io/redhat-appstudio/pull-request-builds:spi-controller-517
quay.io/redhat-appstudio/pull-request-builds:spi-oauth-517
```


pkg/spi-shared/secretstorage/secretstorage.go
```diff
@@ -0,0 +1,48 @@
...
+package secretstorage
+
+import (
+	"context"
+	"errors"
+	"fmt"
+)
+
+// SecretID is a generic identifier of the secret that we store data of. While it very
+// much resembles the Kubernetes client's ObjectKey, we keep it as a separate struct to
```
sparkoo, line 24: `resembles the Kubernetes client's ObjectKey` this is true now, but I think the biggest benefit is to have option to extend it without changing whole `SecretStorage` interface.

pkg/spi-shared/secretstorage/vaultstorage/vault.go
```diff
@@ -106,79 +103,34 @@ type VaultStorageConfig struct {
...
 	}
 
 	return nil
 }
 
-func (v *vaultTokenStorage) Store(ctx context.Context, owner *api.SPIAccessToken, token *api.Token) error {
+func (v *VaultSecretStorage) Store(ctx context.Context, id secretstorage.SecretID, bytes []byte) error {
 	data := map[string]interface{}{
-		"data": token,
+		// yes, the data HAS TO be a map
```
sparkoo, line 124: 😄

pkg/spi-shared/secretstorage/vaultstorage/vault.go
```diff
@@ -192,85 +144,129 @@ func (v *vaultTokenStorage) Store(ctx context.Context, owner *api.SPIAccessToken
...
-	if !dataOk {
-		return nil, fmt.Errorf("%w at '%s'", corruptedDataError, path)
-	}
 
-	return parseToken(data)
-}
-
-func parseToken(data interface{}) (*api.Token, error) {
-	dataMap, ok := data.(map[string]interface{})
+	dataField, ok := secret.Data["data"]
```
sparkoo, line 165: can we keep this in a function and test?

sparkoo (commented):
I think cli packages `awscli` and `vaultcli` has to be moved under secretstorage implementation packages as well. They are not config options for TokenStorage anymore.

Question is what to do with functions `CreateVaultStorage` and `NewAwsTokenStorage`. I think putting them in `tokenstorage` package itself is ok and it should be dependency-cycle-free. Something like `pkg/spi-shared/tokenstorage/factory.go`?

I think it is ok that `tokenstorage` package will depend on `secretstorage` it's implementations. Decoupling it more would be an overkill atm.

WDYT?

mshaposhnik (approved):
looks good at my level of unerstanding of the topic...

sparkoo (commented):
nice work, thanks

may I ask you next time to not reorder functions in a big PR like this? Not that I don't agree with the changes, but it's unreviewable. See `aws.go` file. You can't see what has really changed. Doing just reorder in the followup or preceding PR would be much nicer and safer to review.

And maybe we could talk about some conventions we want to follow how to order functions in a file.

pkg/spi-shared/secretstorage/vaultstorage/vault.go
```diff
@@ -0,0 +1,356 @@
...
+
+	bytes, legacy, err := extractData(secret.Data)
+	if err != nil {
+		return nil, fmt.Errorf("failed to extract the data from Vault response: %w", err)
+	}
+
+	if legacy {
+		// we're dealing with legacy data. Let's convert right now to the new format so that we do
+		// incremental migration to the new data format..
+		if err := v.Store(ctx, id, bytes); err != nil {
```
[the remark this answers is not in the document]
metlos: The only concurrency problem I see here is if 2 concurrent reconciles tried to update the same path in Vault. That could theoretically result in an error (although I am not sure) and that error would bubble up to the reconcilers.

The binding reconciler would flip the binding to the error phase and return the error to the controller manager that would therefore requeue the reconcile.

The token reconciler also flips the token to the error phase and requeues the reconcile as far as I could tell.

xbaran4 (approved):

PR #526, opened 2023-03-20
introduce spi instance id config option and prefix aws secrets with it

sparkoo:
### What does this PR do?
new cli option `--spi-instance-id` or env `SPIINSTANCEID`. For now, it is used only for AWS secrets.

https://issues.redhat.com/browse/SVPI-367

### How to test this PR?

```
```

1. Deploy with AWS secret storage
2. set `SPIINSTANCEID` at `spi-system/spi-shared-environment-config` to some value
3. create binding and provide the token
4. see secret is stored prefixed with `SPIINSTANCEID`

Screenshot 2023-03-20 at 13 14 01
Screenshot 2023-03-20 at 13 15 16
Screenshot 2023-03-20 at 13 15 23


sparkoo:
> wdyt about adding spiInstanceId (aka clusterid) to
> 
> https://github.com/redhat-appstudio/service-provider-integration-operator/blob/ee065809ae70788b257500b0ac993b493e9b4aaf/pkg/spi-shared/secretstorage/secretstorage.go#L27
> 
> ?

https://github.com/redhat-appstudio/service-provider-integration-operator/pull/527/files here's draft to compare. I don't have strong preference. For some reason, I don't feel like spiinstance belongs to secretID.

wdyt: @skabashnyuk @metlos ?

metlos:
I lean towards not having instance id/cluster id in the `SecretID`... See https://issues.redhat.com/browse/SVPI-371?focusedId=21866318&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-21866318

gazarenkov:
I do not think it is spiinstanceid, I'd say it is rather id of the cluster where secret will be created.
In a case of user workload (where gitops deploy app) it is "worker" cluster which in common case is not the same as "control plane" cluster.

sparkoo:
> We need clusterID to be stored to support multi-cluster, just like we have namespace for single-cluster scenario (now) to prevent possiblity to create SPIAccessToken with the same name and use secret data from other namespace.

This issue is not trying to solve multi-cluster issue. It's currently only about using single AWS credentials with multiple SPI instances.

mshaposhnik (approved):

PR #534, opened 2023-03-22
feat: adding action json key to audit logs based on ADR 6

xbaran4:
### What does this PR do?
Adds `action` json key to audit logs where it makes sense.

Note that there are no specific cases mentioned in the ADR 6 just the following:
image
To read more: https://github.com/redhat-appstudio/book/blob/main/ADR/0006-log-conventions.md
So it depends on the preference for which logs should contain which action.

https://issues.redhat.com/browse/SVPI-412

### How to test this PR?

1. deploy
2. `kubectl apply -f samples/spiaccesscheck.yaml`
3. Look at operator logs, they should contain log entry containing "action": "UPDATE" pair similar to this:
image

```
```


sparkoo (commented):
I don't much understand why did you update these audit logs and not others. And I don't much understand how you chose the actions. Requirements in ADR-6 are not really clear to me

mshaposhnik (approved):

sparkoo (approved):

PR #544, opened 2023-03-29
Support workspace context in OAuth & REST upload 

mshaposhnik:
### What does this PR do?

Allows to make workspace-aware calls to K8S API. 

The main idea of the PR, is to make on-demand K8S clients for user-authenticated requests (which are generally limited to token upload or OAuth operations).

It is introduces the factory-builder, which can provide a three different types of factories of the K8S clients:

- InCluster factory, which is able to produce instances of the clients that works on behalf of SPI SA;
  (it's quite ok to get just one instance of the client from that factory and pass it to all the consumers)
  
- UserAuth factory, which is procudes instances of the client which works on behalf of the user (i.e. using his token for authentification) but it is not the workspace-aware. It is used in workspace-less environments like minikube or intergation tests etc

- WorkspaceAware factory, which is also authenticates on server via user token, but also can set the custom API server URL and workspace path in the requests to the cluster,  by consuming the namespace in the context when new client instance is created.
(it is making REST call to the configurable workspace API endpoint under the hood, and find one to which the passed namespace is belongs)

So instead of getting the _singleton-ish_ client instance, workspace-dependent code now just gets the `WorkspaceAware` factory instance, and constructs new client on-demand with all necessary attributes set.

https://issues.redhat.com/browse/SVPI-420
https://issues.redhat.com/browse/SVPI-421

### How to test this PR?

```
```


oauth/client_factory.go
```diff
@@ -40,9 +48,14 @@ type ClientFactoryConfig struct {
...
-// CreateUserAuthClient creates a new client based on the provided configuration. We use this client for k8s requests with user's token (e.g.: SelfSubjectAccessReview).
+type ClientFactory struct {
+	FactoryConfig ClientFactoryConfig
+	HTTPClient    rest.HTTPClient
+}
+
+// CreateUserAuthClientForNamespace creates a new client based on the provided configuration. We use this client for k8s requests with user's token (e.g.: SelfSubjectAccessReview).
 // Note that configuration is potentially modified during the call.
-func CreateUserAuthClient(args *ClientFactoryConfig) (AuthenticatingClient, error) {
+func (f ClientFactory) CreateUserAuthClientForNamespace(ctx context.Context, namespace string) (AuthenticatingClient, error) {
```
metlos, line 60: What about caching the clients? If I remember correctly the Kubernetes client contains an internal object cache which we are probably re-creating each time we create a new client.
We should probably cache per workspace, not per namespace though, because that would mean that we'd have several clients for the same cluster, each having its own object cache.

oauth/context_helper.go
```diff
@@ -0,0 +1,32 @@
...
+import "context"
+
+type contextKey string
+
+func (c contextKey) String() string {
+	return string(c)
+}
+
+var (
+	NamespaceContextKey = contextKey("namespace")
```
metlos, line 26: I would mimic what the `log` package is doing here and provide both `NamespaceFromContext` and `NamespaceIntoContext` functions. This would make it possible to keep the key private and be just an opaque type. 

I think `type contextKey struct {}` and `var namespaceContextKey = contextKey{}` would work fine without it needing to have any concrete value. I would also maybe rename it to something more concrete like `namespaceContextKeyType` just to prevent name collisions if we add some other context key in the future.

oauth/client_factory.go
```diff
@@ -28,6 +29,11 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/log"
 )
 
+var (
+	errUnparseableResponse = errors.New("unable to parse workspace  API response for namespace")
```
metlos: The name `errUnparseableResponse` doesn't correspond with the actual text of the error message too much, IMHO.

metlos: that still doesn't correspond to the text of the error though :) The error message is specific about what kind of request causes this error, yet the name of the error suggests that it is generic for all responses. What about `errUnableToParseWorkspaceResponse`?

oauth/client_factory.go
```diff
@@ -14,156 +14,133 @@
...
+type UserAuthK8sClientFactory struct {
+	ClientOptions *client.Options
+	RestConfig    *rest.Config
 }
 
-func createClient(args *ClientFactoryConfig, mapper meta.RESTMapper, userAuthentication bool) (client.Client, error) {
-	kubeconfig, err := baseKubernetesConfig(args, userAuthentication)
-	if err != nil {
-		return nil, fmt.Errorf("failed to create kubernetes incluster config: %w", err)
+func (w WorkspaceAwareK8sClientFactory) CreateClient(ctx context.Context) (client.Client, error) {
```
sparkoo, line 60: I think it's time to split this into more files, even create new package for client factory

sparkoo: can you please split this very long and complex function? it's doing like 10 different things

cmd/oauth/oauthcli/cli.go
```diff
@@ -21,10 +21,11 @@ import (
...
+	ServiceAddr            string `arg:"--service-addr, env" default:"0.0.0.0:8000" help:"Service address to listen on"`
+	AllowedOrigins         string `arg:"--allowed-origins, env" default:"https://console.redhat.com,https://console.stage.redhat.com,https://console.dev.redhat.com,https://prod.foo.redhat.com:1337" help:"Comma-separated list of domains allowed for cross-domain requests"`
+	KubeConfig             string `arg:"--kubeconfig, env" default:"" help:""`
+	KubeInsecureTLS        bool   `arg:"--kube-insecure-tls, env" default:"false" help:"Whether is allowed or not insecure kubernetes tls connection."`
+	ApiServer              string `arg:"--api-server, env:API_SERVER" default:"" help:"host:port of the Kubernetes API server to use when handling HTTP requests"`
+	ApiServerCAPath        string `arg:"--ca-path, env:API_SERVER_CA_PATH" default:"" help:"the path to the CA certificate to use when connecting to the Kubernetes API server"`
+	ApiServerWorkspacePath string `arg:"--workspace-api-path, env:API_SERVER_WORKSPACE_PATH" default:"/apis/toolchain.dev.openshift.com/v1alpha1/workspaces" help:"path of the workspace API to use when configuring the user K8S clients"`
```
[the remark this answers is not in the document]
mshaposhnik: so the proposal is to make it hardcoded in a constant ?

oauth/client_factory.go
```diff
@@ -14,156 +14,133 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/log"
 )
 
-// AuthenticatingClient is just a typedef that advertises that it is safe to use the WithAuthIntoContext or
-// WithAuthFromRequestIntoContext functions with clients having this type.
-type AuthenticatingClient client.Client
+var (
+	errUnparseableResponse  = errors.New("unable to parse workspace  API response for namespace")
+	errWorkspaceNotFound    = errors.New("target workspace not found for namespace")
+	errClientInstanceNotSet = errors.New("client instance does not set")
```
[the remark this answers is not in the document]
mshaposhnik: this is for test-only factory which handles the instance

oauth/client_factory.go
mshaposhnik: it looks huge due to excessive errors processing, but generally just does the HTTP request and parses the response.
Not sure if a bunch of non-reusable fucntions will improve the situation much... how do you see it better ?

oauth/client_factory.go
mshaposhnik: ok

oauth/client_factory.go
mshaposhnik: I may propose externalize the HTTP stuff, but i'm not sure if it makes sense to make it more granular insede that.

sparkoo: this file has interface and 4 structs implementing this interface. To me, this is already a reason to organize it somehow better than all in one file.

oauth/client_factory.go
```diff
@@ -14,156 +14,139 @@
...
+type InClusterK8sClientFactory struct {
+	ClientOptions *client.Options
+	RestConfig    *rest.Config
 }
 
-// CreateInClusterClient creates a new client based on the provided configuration. We use this client for k8s requests with ServiceAccount (e.g.: reading configuration secrets).
-func CreateInClusterClient(args *ClientFactoryConfig) (client.Client, error) {
-	mapper := meta.NewDefaultRESTMapper([]schema.GroupVersion{})
-	mapper.Add(corev1.SchemeGroupVersion.WithKind("Secret"), meta.RESTScopeNamespace)
+type WorkspaceAwareK8sClientFactory struct {
```
sparkoo, line 48: can you please document what is it doing? Workspaces are not common Kubernetes feature, so it's really useful to have some docs around.

oauth/client_factory.go
```diff
@@ -14,156 +14,139 @@
...
+	if c.client != nil {
+		return c.client, nil
+	} else {
+		return nil, errClientInstanceNotSet
 	}
+}
 
-	if err := v1beta1.AddToScheme(options.Scheme); err != nil {
-		return nil, fmt.Errorf("failed to add api to the scheme: %w", err)
+func fetchWorkspaceName(ctx context.Context, apiServer, namespace string, client rest.HTTPClient) (string, error) {
```
sparkoo, line 109: can you please write docs?

oauth/client_factory.go
```diff
@@ -14,156 +14,139 @@
...
-				return nil, fmt.Errorf("expected to load root CA config from %s, but got err: %w", args.ApiServerCAPath, err)
-			} else {
-				tlsConfig.CAFile = args.ApiServerCAPath
+	wsList := &v1alpha1.WorkspaceList{}
+	if err := json.Unmarshal(bodyBytes, wsList); err != nil {
+		return "", fmt.Errorf("%w '%s': %s", errUnableToParseWorkspaceResponse, namespace, err.Error())
+	}
+	for _, ws := range wsList.Items {
+		for _, ns := range ws.Status.Namespaces {
+			if ns.Name == namespace {
```
sparkoo, line 146: can you have same namespace name in multiple workspaces?

oauth/client_factory.go
```diff
@@ -14,156 +14,139 @@
...
-			// rest.InClusterConfig is doing this most possibly only for early error handling so let's do the same
-			if _, err := certutil.NewPool(args.ApiServerCAPath); err != nil {
-				return nil, fmt.Errorf("expected to load root CA config from %s, but got err: %w", args.ApiServerCAPath, err)
-			} else {
-				tlsConfig.CAFile = args.ApiServerCAPath
+	wsList := &v1alpha1.WorkspaceList{}
+	if err := json.Unmarshal(bodyBytes, wsList); err != nil {
+		return "", fmt.Errorf("%w '%s': %s", errUnableToParseWorkspaceResponse, namespace, err.Error())
+	}
+	for _, ws := range wsList.Items {
```
sparkoo, line 144: what about moving all the fetching of workspaces to own function and do the finding after? It's a a bit lost here with all the mess around fetching, parsing etc. Sth like:
```
workspaces, err := fetchWorkspaces(...)
if err != nil ...

for _, ws := range workspaces {
...
}
```

oauth/client_factory_builder.go
```diff
@@ -0,0 +1,148 @@
...
+	mapper := meta.NewDefaultRESTMapper([]schema.GroupVersion{})
+	mapper.Add(authz.SchemeGroupVersion.WithKind("SelfSubjectAccessReview"), meta.RESTScopeRoot)
+	mapper.Add(v1beta1.GroupVersion.WithKind("SPIAccessToken"), meta.RESTScopeNamespace)
+	mapper.Add(v1beta1.GroupVersion.WithKind("SPIAccessTokenDataUpdate"), meta.RESTScopeNamespace)
+	clientOptions, errClientOptions := clientOptions(mapper)
+	if errClientOptions != nil {
+		return nil, errClientOptions
+	}
+
+	if r.Args.ApiServer == "" {
```
sparkoo, line 80: Does it mean that if I have apiserver set, the cluster must have workspaces available?

sparkoo (commented):
how can I test it?

oauth/client_factory.go
mshaposhnik: So far, it's 1=1. But theoretically, workspace may consist of many namespaces.  And i am not yet heard if the one namespace may join multiple workspaces, that will anyway require some more refactorings.

oauth/client_factory_builder.go
mshaposhnik: at this moment - yes.. is something may be wrong on that ?

oauth/client_factory.go
mshaposhnik: ok

oauth/client_factory.go
mshaposhnik: yes

oauth/client_factory.go
mshaposhnik: ok i have it split per-type

oauth/clientfactory/workspace_aware_client_factory.go
```diff
@@ -0,0 +1,122 @@
...
+
+// fetchWorkspaceName finds the workspace to which the given namespace is belongs
+func fetchWorkspaceName(ctx context.Context, apiServer, namespace string, httpClient rest.HTTPClient) (string, error) {
+	wsList, err := fetchAllAvailableWorkspaces(ctx, apiServer, namespace, httpClient)
+	if err != nil {
+		return "", fmt.Errorf("unable to list available workspaces foe namespace %s, error:%w", namespace, err)
+	}
+	for _, ws := range wsList.Items {
+		for _, ns := range ws.Status.Namespaces {
+			if ns.Name == namespace {
```
sparkoo, line 77: so how workspaces works? Can for example every workspace have `default` namespace? How will this behave then? Does this `WorkspaceAwareK8sClientFactory` expect some exact use of workspaces? Is it tested?

oauth/client_factory_builder.go
sparkoo: it's changing the SPI expectation of cluster setup, it is not documented anywhere that when you set apiserver, you must setup workspaces as well.

oauth/clientfactory/workspace_aware_client_factory.go
mshaposhnik: i would be glad if someone answer those questions. so far nobody knows.

cmd/oauth/oauth.go
```diff
@@ -91,10 +95,15 @@ func main() {
...
+	if errK8sClient != nil {
+		setupLog.Error(errK8sClient, "failed to create user auth k8s client factory")
+		os.Exit(1)
+	}
 	tokenUploader := oauth.SpiTokenUploader{
-		K8sClient: userAuthClient,
+		ClientFactory: userAuthK8sClientFactory,
 		Storage: tokenstorage.NotifyingTokenStorage{
-			Client:       userAuthClient,
+			Client:       inClusterK8sClient,
```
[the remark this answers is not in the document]
mshaposhnik: Yes, that makes sense. I think i'll try to get rid of InCluster client  on OAuth at all.

oauth/clientfactory/workspace_aware_client_factory.go
sparkoo: yes, but what if 2 workspaces have namespace with same name? how do we know which one to use ?

sparkoo: ok, can we document it somehow? It is a lot of internal information from that we're making assumptions in the code.

sparkoo:
> > how can I test it?
> 
> @sparkoo It is something we need to explore. As I understand the plan is:
> 
> * make sure that without API URL everything works as before.
> * Merge this pr and deploy on staging.
> * Check that workspace working correctly. If not revert it.
> * Document existing behavior.
> 
> @mshaposhnik right?

ok, if documentation comes after, I'm ok. I'm just worried that this PR is changing behavior of some configuration (`apiserver`) and is making new assumptions that are not clear to me (`workspaces`).

sparkoo (approved):

PR #548, opened 2023-03-31
Make the TokenUploadReconciler behave more in-line with other reconcilers

metlos:
### What does this PR do?
Make the TokenUploadReconciler behave more in-line with other reconcilers in our codebase.

Most importantly - reconcile the secrets one at a time, which also enables slightly more clear error handling.

none

### How to test this PR?
No change in behavior should be apparent.

```
```


controllers/tokenupload_controller.go
```diff
@@ -61,100 +62,101 @@ type TokenUploadReconciler struct {
...
+	uploadSecret := &corev1.Secret{}
+
+	if err := r.Get(ctx, req.NamespacedName, uploadSecret); err != nil {
+		if errors.IsNotFound(err) {
+			lg.V(logs.DebugLevel).Info("upload secret already gone from the cluster. skipping reconciliation")
+		}
+		return ctrl.Result{}, fmt.Errorf("failed to load the upload secret from the cluster: %w", err)
+	}
+
+	if uploadSecret.DeletionTimestamp != nil {
```
gazarenkov, line 78: For this case it would be rather good to make it deleted ASAP, just immediately after reading ... unfortunately it wont happen :) (not sufficient)

controllers/tokenupload_controller.go
```diff
@@ -163,35 +165,35 @@ func (r *TokenUploadReconciler) logError(ctx context.Context, secret corev1.Secr
...
 	err = r.Create(ctx, secretErrEvent)
 	if err != nil {
-		lg.Error(err, "Event creation failed for Secret: ", "secret.name", secret.Name)
+		lg.Error(err, "event creation failed for upload secret")
 	}
 
 }
 
-// Contract: having at only one event if upload failed and no events if uploaded.
+// Contract: having at most one event if upload failed and no events if uploaded.
```
gazarenkov, line 181: It should be exactly one actually

gazarenkov (approved):

controllers/tokenupload_controller.go
metlos: We cannot guarantee that the reconciler won't be triggered between the time the secret is marked for deletion and actually deleted. This check is always there just to bail out quickly in a situation where the secret is in the process of being deleted (but the process is waiting for some condition that is outside of our responsibility).

sparkoo (approved):

PR #549, opened 2023-03-31
Clean up the tokenstorage/secretstorage split and introduce TypedSecretStorage

metlos:
### What does this PR do?
Clean up the tokenstorage/secretstorage split and introduce TypedSecretStorage to aid in adding new storage types in the future.

none

### How to test this PR?
no changes in behavior should be apparent.
```
```

### DIFF
--- a/pkg/spi-shared/secretstorage/secretstorage.go
+++ b/pkg/spi-shared/secretstorage/secretstorage.go
@@ -16,8 +16,11 @@ package secretstorage
 
 import (
 	"context"
+	"encoding/json"
 	"errors"
 	"fmt"
+
+	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
 // SecretID is a generic identifier of the secret that we store data of. While it very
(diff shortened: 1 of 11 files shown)


sparkoo:
Do you need this for remote secrets? will it use different secretstorage type? Or what problem you're trying to solve with this PR?

metlos:
> Do you need this for remote secrets? will it use different secretstorage type? Or what problem are you trying to solve with this PR?

Yes, remote secrets will be storing different type of data and will be keyed by a different object type. But the CRUD logic will be the same, so I wondered why not provide a generic impl to which the, to be created, `RemoteSecretStorage` would basically be just a type alias.

pkg/spi-shared/secretstorage/secretstorage.go
```diff
@@ -47,3 +50,112 @@ type SecretStorage interface {
...
+	// Get retrieves the data under the given id. A NotFoundError is returned if the data is not found.
+	Get(ctx context.Context, id *ID) (*D, error)
+	// Delete deletes the data of given id. A NotFoundError is returned if there is no such data.
+	Delete(ctx context.Context, id *ID) error
+}
+
+// DefaultTypedSecretStorate is the default implementation of the TypedSecretStorage interface
+// that uses the provided functions to convert between the id and data types to SecretID and []byte
+// respectively.
+type DefaultTypedSecretStorate[ID any, D any] struct {
```
sparkoo, line 70: can you please write tests for `DefaultTypedSecretStorate`

sparkoo: also I'm not sure about the name `Default`, but we can rename later when we have more implementations and know more

pkg/spi-shared/tokenstorage/aws.go
```diff
@@ -0,0 +1,31 @@
...
+package tokenstorage
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/redhat-appstudio/service-provider-integration-operator/pkg/spi-shared/secretstorage/awsstorage/awscli"
+)
+
+func NewAwsTokenStorage(ctx context.Context, args *awscli.AWSCliArgs) (TokenStorage, error) {
```
sparkoo, line 24: this is bringing `awscli` dependency into `tokenstorage` package. I'm not sure it's ok. I have some feeling I wanted to avoid that, but I don't remember exactly why :D

sparkoo (approved):

PR #555, opened 2023-04-06
fix for vault-init script with realpath error

xbaran4:
### What does this PR do?
Currently, this PR has 3 changes in the vault-init.sh script:
1. L49: This hides the error from k8s API server when the `spi-vault-keys` secret does not exist. This error is part of an if statement and is expected therefore it should not be printed out from our script which may seem like our script is failing. 
2. L163: This fixes the actual problem with realpath on Mac. When the `approle_secret.yaml` file does not exist yet, the Mac realpath fails with `realpath: .tmp/approle_secret.yaml: No such file or directory`. This does not happen on Linux. To prevent this, we are concatenating realpath of the `.tmp` directory, which must exist due to the command above, with the yaml file name.
3. L204: Changes the message to be more specific about what exact state we are expecting the pod to be in.

### How to test this PR?

```
```


mshaposhnik (approved):

hack/vault-init.sh
```diff
@@ -160,7 +160,7 @@ function approleAuth() {
   fi
 
   if [ ! -d ".tmp" ]; then mkdir -p .tmp; fi
-  SECRET_FILE=$( realpath .tmp/approle_secret.yaml )
+  SECRET_FILE="$( realpath .tmp)/approle_secret.yaml"
```
sparkoo, line 163:
```suggestion
  SECRET_FILE="$( realpath .tmp )/approle_secret.yaml"
```

sparkoo (approved):

PR #563, opened 2023-04-11
RemoteSecret MVP, take 2

metlos:
### What does this PR do?

This introduces an abstraction over how we deliver the secret and service accounts such that we can support both the current "target" to the namespace of the binding as well as other targets, like other namespace or an environment.

This is a little bit invasive refactoring so, some unit tests are commented out for the time being. Note though, that the integration tests are still working, providing a proof that the refactoring is able to at least support the current functionality :)

In order to lower the havoc this has caused in the codebase, I've kept the "controllers/bindings" package as the "root" that defines how the delivery to a target is done, even though it can now potentially deliver to other target than just the bindings. We should rename this package to something like "dependents" or "delivery" or something.

To understand how this is all done, look at the interplay of the interfaces defined in the `controllers/bindings/targets.go` and the functionality defined in `controllers/bindings/dependents.go`.

https://issues.redhat.com/browse/SVPI-391

### How to test this PR?
N/A as this is just a draft of an MVP.

```
```


api/v1beta1/remotesecret_types.go
```diff
@@ -0,0 +1,80 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// RemoteSecretSpec defines the desired state of RemoteSecret
+type RemoteSecretSpec struct {
+	Secret LinkableSecretSpec `json:"secret"`
+	Target RemoteSecretTarget `json:"target"`
+}
+
+type RemoteSecretTarget struct {
```
gazarenkov, line 29: What is Environment?
A name of the secret with kubeconfig for connection to other cluster?

controllers/bindings/targets.go
```diff
@@ -0,0 +1,143 @@
...
+	"context"
+
+	api "github.com/redhat-appstudio/service-provider-integration-operator/api/v1beta1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// SecretDeploymentTarget, together with SecretBuilder and ObjectMarker, represents a method of obtaining enough
+// information for the DependentsHandler to be able to deliver the secrets and service accounts to some "target"
+// place in (some) K8s cluster.
+type SecretDeploymentTarget interface {
```
gazarenkov, line 27: Why do we need it as an interface?
My understanding it is considered to have 3 implementations (same-namespace, other-namespace, other-cluster) ?
First 2 differs by namespace name, third - (will) have other client... Or I misunderstand something?

api/v1beta1/remotesecret_types.go
metlos: Environment is the appstudio environment. this is being talked about as our next step in developing this functionality.

controllers/bindings/targets.go
metlos: The interface hides the different mechanics of obtaining the namespace and the client.

api/v1beta1/remotesecret_types.go
gazarenkov: I see.
I would not like we have direct dependencies on appstudio logical entities here. It is all about physical location (namespace + (maybe) cluster), let external entities care about this mapping, not RemoteSecret.

controllers/bindings/targets.go
gazarenkov: I'd avoid doing that in the core model assuming that business logic on top can manage this mapping, we Probably can imagine it as a pluggable mechanism, but let's start simply.

controllers/spiaccesstokenbinding_controller.go
mshaposhnik, line 1: Set L132: 

```
return binding.Status.LinkedAccessTokenName == dataUpdate.Spec.DataOwner.Name
```

that will fix the test

controllers/remotesecrets/secretbuilder.go
```diff
@@ -0,0 +1,45 @@
...
+	"errors"
+	"fmt"
+
+	api "github.com/redhat-appstudio/service-provider-integration-operator/api/v1beta1"
+	"github.com/redhat-appstudio/service-provider-integration-operator/controllers/bindings"
+	"github.com/redhat-appstudio/service-provider-integration-operator/pkg/spi-shared/remotesecretstorage"
+	"github.com/redhat-appstudio/service-provider-integration-operator/pkg/spi-shared/secretstorage"
+)
+
+type SecretBuilder struct {
```
sparkoo, line 28: I'm lost in those `SecretBuilder`s. So there is `SecretBuilder` interface and 2x `SecretBuilder` implementations of that interface? All with same name? How are they different? What are they doing? Can you write short doc, maybe name them somehow so it's more clear what is it doing?

config/crd/kustomization.yaml
```diff
@@ -7,4 +7,5 @@ resources:
 - bases/appstudio.redhat.com_spiaccesstokendataupdates.yaml
 - bases/appstudio.redhat.com_spiaccesschecks.yaml
 - bases/appstudio.redhat.com_spifilecontentrequests.yaml
+- bases/appstudio.redhat.com_remotesecrets.yaml
```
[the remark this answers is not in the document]
metlos: yes

controllers/spiaccesstokenbinding_controller.go
metlos: This was only a part of the problem that should now be fixed in the latest commit.

controllers/bindings/targets.go
```diff
@@ -0,0 +1,143 @@
...
+	// GetClient returns the client to use when connecting to the target "destination" to deploy the dependent objects to.
+	GetClient() client.Client
+	// GetType returns the type of the secret deployment target object.
+	GetType() string
+	// GetTargetObjectKey is the location of the object that describes the target. This is mainly used in error reporting.
+	GetTargetObjectKey() client.ObjectKey
```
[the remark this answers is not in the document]
metlos: you may want to read the comments.

controllers/bindingtarget/objectmarker.go
```diff
@@ -0,0 +1,145 @@
...
+
+	return changed, nil
+}
+
+// MarkReferenced implements dependents.ObjectMarker
+func (m *BindingTargetObjectMarker) MarkReferenced(ctx context.Context, binding client.ObjectKey, obj client.Object) (bool, error) {
+	changed := false
+	annos := obj.GetAnnotations()
+
+	val := commaseparated.Value(annos[LinkAnnotation])
```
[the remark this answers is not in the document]
metlos: this has been the case for the last couple of months. A quick google search didn't reveal a max size of annotation value to me but i'm not too worried since for example kubectl stores the last applied value of the whole object in its own annotation.

controllers/namespacetarget/objectmarker.go
```diff
@@ -0,0 +1,189 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+const (
+	LinkedByRemoteSecretLabel          = "spi.appstudio.redhat.com/linked-by-remote-secret" //#nosec G101 -- false positive, this is just a label
+	ManagingRemoteSecretNameAnnotation = "spi.appstudio.redhat.com/managing-remote-secret"  //#nosec G101 -- false positive, this is just a label
+	LinkedRemoteSecretsAnnotation      = "spi.appstudio.redhat.com/linked-remote-secrets"   //#nosec G101 -- false positive, this is just a label
+)
+
+type NamespaceObjectMarker struct {
```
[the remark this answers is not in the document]
metlos: the name says it all.

controllers/remotesecrets/secretbuilder.go
metlos: you may have noticed that names in Go are always namespaced (unless you are doing a * import which is discouraged in most cases). I would argue that stuff like `tokenstorage.TokenStorage` is stylistically worse than `remotesecrets.SecretBuilder`.

metlos: That said, I do agree that `SecretBuilder` is a bad name for this interface and have renamed it to an uninspired but descriptive `SecretDataGetter`.

controllers/starter.go
```diff
@@ -90,20 +106,47 @@ func SetupAllReconcilers(mgr controllerruntime.Manager, cfg *config.OperatorConf
 		return err
 	}
 
+	if err = (&RemoteSecretReconciler{
```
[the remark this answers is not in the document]
metlos: Good point. Feature flag added.

controllers/bindings/targets.go
```diff
@@ -0,0 +1,143 @@
...
+	// GetActualSecretName returns the actual name of the secret, if any (as opposed to the
+	// configured name from the spec, which may not fully represent what's in the cluster
+	// if for example GenerateName is used).
+	GetActualSecretName() string
+	// GetActualServiceAccountNames returns the names of the service accounts that the spec
+	// configures.
+	GetActualServiceAccountNames() []string
+}
+
+// SecretBuilder is an abstraction that, given the provided key, is able to obtain the secret data from some kind of backing
```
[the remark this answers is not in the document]
metlos: renamed to `SecretDataGetter`.

controllers/bindings/dependents.go
```diff
@@ -21,16 +21,15 @@ import (
 	"github.com/cenkalti/backoff/v4"
 	api "github.com/redhat-appstudio/service-provider-integration-operator/api/v1beta1"
 	"github.com/redhat-appstudio/service-provider-integration-operator/pkg/serviceprovider"
-	"github.com/redhat-appstudio/service-provider-integration-operator/pkg/spi-shared/tokenstorage"
 	corev1 "k8s.io/api/core/v1"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
 // DependentsHandler is taking care of the dependent objects of the binding
```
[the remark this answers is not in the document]
metlos: I've gone through the code and removed the mentions of "binding" in the places that are now target agnostic.

api/v1beta1/remotesecret_types.go
metlos: yes, agreed. I removed the hints of the environment from the code.

controllers/bindings/errors.go
```diff
@@ -16,6 +16,23 @@ package bindings
...
+const (
+	ErrorReasonNone ErrorReason = ""
+
+	// XXX: note that this used to be used as:
+	// - api.SPIAccessTokenBindingErrorReasonTokenSync originally in secretHandler.Sync
+	ErrorReasonSecretUpdate ErrorReason = "SecretUpdate"
+	// XXX: note that this used to be used as:
+	// - api.SPIAccessTokenBindingErrorReasonServiceAccountUnavailable in ensureReferencedServiceAccount -> serviceAccountHandler.Sync
+	ErrorReasonServiceAccountUnavailable ErrorReason = "ServiceAccountUnavailable"
+	// XXX: note that this used to be used as:
```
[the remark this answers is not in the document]
metlos: https://www.oracle.com/java/technologies/javase/codeconventions-programmingpractices.html#395

controllers/bindings/errors.go
```diff
@@ -16,6 +16,23 @@ package bindings
...
+	// - api.SPIAccessTokenBindingErrorReasonServiceAccountUpdate in ensureReferencedServiceAccount -> serviceAccountHandler.Sync
+	// - api.SPIAccessTokenBindingErrorReasonTokenSync in ensureReferencedServiceAccount -> serviceAccountHandler.Sync
```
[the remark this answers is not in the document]
metlos: We always took great care to list all the error reasons as the values of the appropriate type in the CRD. In this MVP I forgave that in the name of simplicity and am simply forcefully converting a string into the error reason type mentioned in the CRD. I think it would be good to go back to using the strongly typed error reasons and therefore left a note here in the code.

To answer your supposed follow up question with a question... Is it right to create an issue for fixing something that is not yet in the main branch?

cmd/cmd.go
```diff
@@ -19,38 +19,36 @@ import (
...
-	errUnsupportedTokenStorage = errors.New("unsupported token storage type")
-	errNilTokenStorage         = errors.New("nil token storage")
+	errUnsupportedSecretStorage = errors.New("unsupported secret storage type")
+	errNilSecretStorage         = errors.New("nil secret storage")
 )
 
-func InitTokenStorage(ctx context.Context, args *CommonCliArgs) (tokenstorage.TokenStorage, error) {
-	var tokenStorage tokenstorage.TokenStorage
-	var errTokenStorage error
+func UninitializedSecretStorage(ctx context.Context, args *CommonCliArgs) (secretstorage.SecretStorage, error) {
```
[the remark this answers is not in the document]
metlos: maybe not

hack/remote-secret-secret.yaml
```diff
@@ -0,0 +1,15 @@
+apiVersion: v1
+kind: Secret
```
sparkoo, line 2: would it be better to have this in `samples` dir? Can you please also create some `RemoteSecret` sample yaml?

sparkoo:
testing with AWS, I'm seeing tokenstorage initialized 4times. Only in operator

```
2023-04-21T06:19:37.056Z    DEBUG    awscli/cli.go:54    creating AWS config    {"spiInstanceId": "spi-1"}                                                                                                                                                                                                                                                                   
2023-04-21T06:19:37.056Z    INFO    awsstorage/aws.go:57    initializing AWS token storage    {"secretstorage": "AWS"}                                                                                                                                                                                                                                                       
2023-04-21T06:19:37.232Z    INFO    awsstorage/aws.go:57    initializing AWS token storage    {"secretstorage": "AWS"}                                                                                                                                                                                                                                                       
2023-04-21T06:19:37.347Z    INFO    awsstorage/aws.go:57    initializing AWS token storage    {"secretstorage": "AWS"}                                                                                                                                                                                                                                                       
2023-04-21T06:19:37.437Z    INFO    awsstorage/aws.go:57    initializing AWS token storage    {"secretstorage": "AWS"}                                                                                                                                                                                                                                                       
2023-04-21T06:19:37.529Z    INFO    setup    operator/operator.go:132    starting manager
```

sparkoo:
AWS is not working because the instance id is not configured properly. instance id is set in context https://github.com/redhat-appstudio/service-provider-integration-operator/blob/main/cmd/operator/operator.go#L96 and tokenstorage was initialized with same context. This PR is changing initialization process and it ends with different context so instance id is not there.

hack/remote-secret-secret.yaml
metlos: good catch

sparkoo (approved):
I've tested some basic scenarios with AWS secretsmanager, everything I've tested works as expected.

few issues I've noticed, not blocking the merge, but something to think of as followup given the circumstances:
1. tokenstorage is initialized twice:
```
2023-04-21T13:54:50.633Z    INFO    awsstorage/aws.go:58    initializing AWS token storage    {"spiInstanceId": "spi-1", "secretstorage": "AWS"}                                                                                                                                                                                                                             
2023-04-21T13:54:50.823Z    INFO    awsstorage/aws.go:58    initializing AWS token storage    {"spiInstanceId": "spi-1", "secretstorage": "AWS"} 
```

2. some cleanup work in storage area (I've created issue https://issues.redhat.com/browse/SVPI-424)
3. Only way to upload remotesecret is upload with secret, but it is possible to configure SPI to enable remotesecret controller and disable secretupload controller, which does not make much sense
4. test coverage dropped a lot

cmd/oauth/oauth.go
```diff
@@ -89,8 +89,20 @@ func main() {
...
-	strg, err := cmd.InitTokenStorage(ctx, &args.CommonCliArgs)
+	storage, err := cmd.CreateInitializedSecretStorage(ctx, &args.CommonCliArgs)
 	if err != nil {
+		setupLog.Error(err, "failed to construct the secret storage")
+		os.Exit(1)
+	}
+	if err := storage.Initialize(ctx); err != nil {
+		setupLog.Error(err, "failed to initialize the secret storage")
+		os.Exit(1)
+	}
```
xbaran4, line 100: Is the `Initialized` in CreateInitializedSecretStorage() meant differently than the `storage.Initialize()`? Because I would assume that the storage created by CreateInitializedSecretStorage does not need to be further initialized.

mshaposhnik (approved):

cmd/oauth/oauth.go
sparkoo: 🫨

pkg/spi-shared/commaseparated/commaseparated.go
xbaran4, line 1: Correct me if I am wrong, but as far as I understand, this is just a standard SET of strings that can be constructed from or turned into a "master" string where all the values of the SET are separated by a comma. 

Isn't it better to implement this through some library such as https://github.com/deckarep/golang-set (the first one I found :D I don't know the quality) or through Go's map rather than having this somewhat complex logic here?

controllers/tokens/secretdatagetter.go
```diff
@@ -0,0 +1,66 @@
...
+	}
+
+	stringData, err := at.ToSecretType(&sb.Binding.Spec)
+	if err != nil {
+		return nil, string(api.SPIAccessTokenBindingErrorReasonTokenAnalysis), fmt.Errorf("failed to create data to be injected into the secret: %w", err)
+	}
+
+	// copy the string data into the byte-array data so that sync works reliably. If we didn't sync, we could have just
+	// used the Secret.StringData, but Sync gives us other goodies.
+	// So let's bite the bullet and convert manually here.
```
xbaran4, line 57: Do you know if this comment about sync is still relevant?

xbaran4: Ah, I see this is called from the sync, so it is relevant.

pkg/spi-shared/commaseparated/commaseparated.go
metlos: Yes, this is just a set with a nice API. `commaseparated.Value("a,b").Contains("c")` reads much nicer than doing it manually IMHO. As for using a library. Yes, we probably could, but at the same time, this is 99 lines of code including the 12 lines of license header. It also has 100% test coverage.

pkg/spi-shared/secretstorage/vaultstorage/vault.go
```diff
@@ -255,11 +255,15 @@ func (v *VaultSecretStorage) login(ctx context.Context) error {
 func (v *VaultSecretStorage) initMetrics(ctx context.Context) error {
 	if v.Config.MetricsRegisterer != nil {
 		if err := v.Config.MetricsRegisterer.Register(vaultRequestCountMetric); err != nil {
-			return fmt.Errorf("failed to register request count metric: %w", err)
+			if !errors.As(err, &prometheus.AlreadyRegisteredError{}) {
```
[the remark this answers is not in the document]
metlos: I added this check to prevent a hard exit in case we accidentally called the `Initialize` of vault storage more times. That used to happen at one point in evolution of the PR, but shouldn't anymore. That said, a hard exit is a bit harsh if it is avoidable in an easy way like this, it should, so I left the check in.

controllers/tokens/secretdatagetter.go
metlos: good catch, no need for that comment anymore because we need to convert due to the return type now.

cmd/oauth/oauth.go
metlos: God, I added this on two places and forgot to remove them twice 🙄

controllers/tokenupload_controller.go
```diff
@@ -121,15 +156,40 @@ func (r *TokenUploadReconciler) Reconcile(ctx context.Context, req ctrl.Request)
...
+		return fmt.Errorf("can not find RemoteSecret: %w ", err)
+	} else if remoteSecret == nil {
+		// SPIAccessToken does not exist, so create it
+		remoteSecret, err = r.createRemoteSecret(ctx, uploadSecret, lg)
+		if err != nil {
+			return fmt.Errorf("can not create RemoteSecret: %w ", err)
+		}
+	}
+
+	auditLog := logs.AuditLog(ctx).WithValues("remoteSecretName", remoteSecret.Name)
```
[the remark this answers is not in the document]
metlos: The analogous code for tokens has an `UPDATE` action set, so I did the same. Great catch.

xbaran4 (approved):
I tried to be as thorough as possible, just had some questions otherwise LGTM. Huge amount of work, well done.

cmd/operator/operatorcli/cli.go
```diff
@@ -34,4 +34,5 @@ type OperatorCliArgs struct {
 	DeletionGracePeriod         time.Duration      `arg:"--deletion-grace-period, env" default:"2s" help:"The grace period between a condition for deleting a binding or token is satisfied and the token or binding actually being deleted."`
 	MaxFileDownloadSize         int                `arg:"--max-download-size-bytes, env" default:"2097152" help:"A maximum file size in bytes for file downloading from SCM capabilities supporting providers"`
 	EnableTokenUpload           bool               `arg:"--enable-token-upload, env" default:"true" help:"Enable Token Upload controller. Enabling this will make possible uploading access token with Secrets."`
+	EnableRemoteSecrets         bool               `arg:"--enable-remote-secrets, env" default:"true" help:"Enable the RemoteSecret controller."`
```
sparkoo, line 37:
```suggestion
	EnableRemoteSecrets         bool               `arg:"--enable-remote-secrets, env" default:"false" help:"Enable the RemoteSecret controller."`
```

cmd/cmd_test.go
```diff
@@ -28,25 +28,18 @@ func TestCreateTokenStorage(t *testing.T) {
...
-		strg, err := InitTokenStorage(context.TODO(), &CommonCliArgs{TokenStorage: blabol})
+		ss, err := CreateInitializedSecretStorage(context.TODO(), &CommonCliArgs{TokenStorage: blabol})
 
-		assert.Nil(t, strg)
+		assert.Nil(t, ss)
 		assert.Error(t, err)
-		assert.ErrorIs(t, err, errUnsupportedTokenStorage)
-	})
-
-	t.Run("fail vault init", func(t *testing.T) {
```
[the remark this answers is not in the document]
metlos: I removed it when the `CreateInitializedSecretStorage` was called `UninitializedSecretStorage` and didn't trigger the error condition. Good catch.

controllers/bindings/targets.go
```diff
@@ -0,0 +1,143 @@
...
+	MarkReferenced(ctx context.Context, obj client.Object) (bool, error)
+	UnmarkReferenced(ctx context.Context, obj client.Object) (bool, error)
+	IsManaged(ctx context.Context, obj client.Object) (bool, error)
+	IsManagedByOther(ctx context.Context, obj client.Object) (bool, error)
+	IsReferenced(ctx context.Context, obj client.Object) (bool, error)
+	ListManagedOptions(ctx context.Context) ([]client.ListOption, error)
+	ListReferencedOptions(ctx context.Context) ([]client.ListOption, error)
+}
+
+type TestDeploymentTarget struct {
```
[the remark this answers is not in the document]
metlos: will move it in the upcoming PR with the tests.

PR #564, opened 2023-04-13
Providing the option to specify dockerconfigjson auth format

xbaran4:
### What does this PR do?
Kubernetes' interpretation of docker config.json is more permissive than Docker's. This PR allows users to alter the json we create for Secrets of type kubernetes.io/dockerconfigjson.

Implements https://issues.redhat.com/browse/SVPI-446

### How to test this PR?

1. Deploy and create binding 
```yaml
kind: SPIAccessTokenBinding
metadata:
  name: test-binding-dockerconfigjson
  namespace: default
spec:
  permissions:
    required:
      - type: rw
        area: registry
  repoUrl: quay.io/repository/redhat_emp1/spi-test
  secret:
    type: kubernetes.io/dockerconfigjson
    name: test-secret-dockerconfigjson
    annotations:
      spi.appstudio.redhat.com/config-json-type: kubernetes
```
2. upload secret
3. check the secret created by the binding. The format should be as follows:
```json
{
    "auths": {
        "quay.io/repository/redhat_emp1/spi-test": {
            "auth": "…"
        }
    }
}
```
4. edit the secret's annotations in binding to:
```yaml
annotations:
      spi.appstudio.redhat.com/config-json-type: explicit
      spi.appstudio.redhat.com/config-json-auth-key: my.custom.host/test
```
and check the secret. It should have the format:
```json
{
    "auths": {
        "my.custom.host/test": {
            "auth": "…"
        }
    }
}
```
5. Edit the secret's annotations in binding to:
```yaml
annotations:
      spi.appstudio.redhat.com/config-json-type: docker
```
and check the secret. It should have the format:
```json
{
    "auths": {
        "quay.io": {
            "auth": "…"
        }
    }
}
```
6. Remove all secret annotations from the binding and ensure the format stays the same.

```
```


pkg/serviceprovider/access_token_mapper.go
```diff
@@ -30,6 +30,11 @@ import (
 
 // Key for token using in Opaque Secret
 const tokenKey = "token"
+const (
+	dockerConfigJsonTypeAnnotationKey     = "spi.appstudio.redhat.com/config-json-type"
+	dockerConfigJsonExplicitAnnotationKey = "spi.appstudio.redhat.com/config-json-auth-key"
+	dockerConfigJsonTypeKubernetes        = "kubernetes"
```
metlos, line 36: To be maximally explicit, I would have 3 `dockerConfigJSonType`s:
* `docker` - for setting the host as the auth key, as docker does,
* `kubernetes` - setting the full provider URL as the auth key,
* `explicit` - for setting the auth key using the value in the `spi.appstudio.redhat.com/config-json-auth-key`.

That way, the configuration is explicit and we don't need to talk about the precedence of the values, etc.

metlos (commented):
In general LGTM, but see the suggestion above that IMHO improves the UX ever so slightly.

pkg/serviceprovider/access_token_mapper.go
xbaran4: Good idea.

xbaran4:
> I'm a bit worried that the tag name if it is defined is also included. Is it by design?

Good catch. I don't have a clear answer to that. On one hand, the tag is a valid part of the URL's path, so yes it is in the json by design. On the other hand, It is possible that when Kubernetes looks for credentials to pull the `golang:1.17` it might consider just URL paths ending with `golang` and not `golang:1.17` which would be a bug. I will try to research this and get back to you.

By the way, do you know if SPI behaves differently for repoUrls with a tag or without a tag with the same image?

mshaposhnik (approved):

PR #565, opened 2023-04-14
Finish the RemoteSecretController + a bunch of smallish refactorings to support it.

metlos:
### What does this PR do?
DO NOT MERGE THIS!!!

This is a PR made to a temporary branch that corresponds to the PR #563 to limit the amount of change one needs to look at.

This finishes off the `RemoteSecretController` adding support for automagic cleanup of linked resources, reaction to data updates, not mentioning the actual deployment of the secrets and service accounts to other namespaces.

https://issues.redhat.com/browse/SVPI-391

### How to test this PR?
1. Look at the `hack/remote-secret-secret.yaml` for an example of a secret that is interpreted as a remote secret when handled by the `TokenUploadController`.
2. See that the following works:
  a. Secrets and SAs are deployed to the target namespace (I actually didn't even test a service account, because that is not doable using just the upload secret)
  b. the deployed secrets and SAs are watched for changes
  c. when the `RemoteSecret` CR is deleted, the linked secret and SAs are deleted, too. The data should also be delete from the backing storage.

```
quay.io/lkrejci/service-provider-integration-oauth:namespace-target
quay.io/lkrejci/service-provider-integration-operator:namespace-target
```


metlos:
closing this in favor of https://github.com/redhat-appstudio/service-provider-integration-operator/pull/563 that now contains all the changes..

PR #581, opened 2023-04-25
Improve the URL examining

mshaposhnik:
### What does this PR do?

Improve the URL examining by more accurate URL matching

### How to test this PR?

```
```


pkg/serviceprovider/github/github.go
```diff
@@ -345,8 +347,12 @@ type githubProbe struct{}
...
 var _ serviceprovider.Probe = (*githubProbe)(nil)
 
-func (g githubProbe) Examine(_ *http.Client, url string) (string, error) {
-	if strings.HasPrefix(url, config.ServiceProviderTypeGitHub.DefaultBaseUrl) {
+func (g githubProbe) Examine(_ *http.Client, repoBaseurl string) (string, error) {
+	baseUrl, err := url.Parse(repoBaseurl)
+	if err != nil {
+		return "", fmt.Errorf("%w '%s'", failedToParseRepoUrlError, repoBaseurl)
+	}
+	if fmt.Sprintf("%s://%s", baseUrl.Scheme, baseUrl.Host) == config.ServiceProviderTypeGitHub.DefaultBaseUrl {
```
gazarenkov, line 355: What about subdomains and specifically "www."  ?
My understanding they should fit as well, no?

metlos (approved):

pkg/serviceprovider/github/github.go
mshaposhnik: if it is set in SP configuration, then it will match. i would better avoid any king of guessing here.

sparkoo: how will it match? it's comparing against static `https://github.com`

mshaposhnik: So far the goal is to improve the matching, not yet taking care of how it's configured in the SP. It may be hardcoded or became dynamic soon (@xbaran4 is working on that right now afaik) but it should check it strictly.

metlos: Agreed. The current approach is to use static URLs so let's stick to it here. Pavol is working on improving that, so let's deal with that once we have an established approach of handling per provider URLs. It will most likely conflict with this change, so we're not going to miss this.

PR #590, opened 2023-05-01
ensure binding and filecontentrequest with custom github url match correct token

xbaran4:
### What does this PR do?
When a custom GitHub URL is set in the shared configuration, this PR ensures that:
1. SPIAccessTokenBinding with repoUrl based on that custom URL is linked to a correct SPIAccessToken (token should have GitHub type and custom host. e.g., mygithub.com). 
2. SPIFileContentRequest with repoUrl based on that custom URL correctly parses the owner and repository from that repoUrl.

NOTE: This PR is not enough to make SPI work with GitHub instances with custom URLs (AKA GitHub Enterprise Server).
The reason is we still use the default GitHub client in code (for calling GitHub API), which uses https://github.com URL. I wrote down a comment which explains this situation and what we need to do, I will copy it here:

We need the githubBaseUrl in the githubClientBuilder struct to decide whether to create a regular GitHub client or an Enterprise client.
The github library provides NewEnterpriseClient() method to create Enterprise clients.
However, it takes 2 URLs as parameters: one for the API URL and one for the upload URL where files can be uploaded.
These two URLs are not the same by default. baseURL should be in format http(s)://[hostname]/api/v3/ and
uploadURL should be in this format http(s)://[hostname]/api/uploads/.
Note that it is likely that the owner of the GitHub Enterprise instance can change the URLs. Thus, we should not assume anything.
Our current configuration interfaces only allow for a single URL to be specified.
As far as I know, we currently do not allow uploading files to GitHub, which means we might omit the uploadURL.
However, we should decide formally whether to change the service provider configuration interface to allow for two URLs or omit the uploadURL (or assume that the uploadURL will be in some format).

image
image

Part of https://issues.redhat.com/browse/SVPI-453

### How to test this PR?

 1. Edit the shared service provider configuration (restart deployments afterward) so that GitHub has a custom URL e.g., to:
 ```
serviceProviders:
- type: GitHub
  clientId: "123"
  clientSecret: "42"
  baseUrl: "https://mygithub.com"
```
2. Create binding with repoUrl based on the custom URL:
```bash
kind: SPIAccessTokenBinding
metadata:
  name: test-binding
  namespace: default
spec:
  permissions:
    required:
      - type: rw
        area: repository
  repoUrl: https://mygithub.com/testingorganization/test
  secret:
    name: test-secret
EOF
```
 3. Check the linked SPIAccessToken; it should have the type GitHub, the host should be mygithub.com, and it should have an oauthURL. This checks the first feature of the PR.
 
 1. On regular GitHub (https://github.com/settings/tokens), create a personal access token with `repo` scope and save it.
 2. Create SPIFileContentRequest but change the path to reference your private repository on github.com. The URL mygithub.com is intentional here; the reason is that right now, we only parse the owner and repository name out of the repoUrl but always use a client with github.com URL: 
 ```bash
kind: SPIFileContentRequest
metadata:
  name: test-file-content-request
  namespace: default
spec:
  repoUrl: https://mygithub.com/xbaran4/testing
  filePath: devfile.yaml
EOF
```
3. Check the SPIAccessTokenBinding created for the SPIFileContentRequest; it should have the oauthUrl and uploadUrl.
4. Upload your PAT token for the SPIAccessTokenBinding's SPIAccessToken.
5. If you uploaded a valid token and the repoUrl in SPIFileContentRequest had a valid path to your repo, the SPIFileContentRequest should be in phase: Delivered

```
```

### DIFF
--- a/pkg/serviceprovider/github/downloadfilecapability.go
+++ b/pkg/serviceprovider/github/downloadfilecapability.go
@@ -22,6 +22,8 @@ import (
 	"net/http"
 	"regexp"
 
+	"github.com/redhat-appstudio/service-provider-integration-operator/pkg/logs"
+
 	"github.com/google/go-github/v45/github"
 	api "github.com/redhat-appstudio/service-provider-integration-operator/api/v1beta1"
 	"github.com/redhat-appstudio/service-provider-integration-operator/pkg/serviceprovider"
@@ -31,6 +33,7 @@ import (
 type downloadFileCapability struct {
 	httpClient      *http.Client
 	ghClientBuilder githubClientBuilder
+	githubBaseUrl   string
 }
 
 var _ serviceprovider.DownloadFileCapability = (*downloadFileCapability)(nil)
@@ -42,25 +45,17 @@ var (
 	unexpectedRepoUrlError     = errors.New("repoUrl has unexpected format")
 )
 
-var _URLRegexp = regexp.MustCompile(`(?Um)^(?:https)(?:\:\/\/)github.com/(?P<owner>[^/]+)/(?P<repo>[^/]+)(.git)?$`)
-var _URLRegexpNames = _URLRegexp.SubexpNames()
-
 func (f downloadFileCapability) DownloadFile(ctx context.Context, repoUrl, filepath, ref string, token *api.SPIAccessToken, maxFileSizeLimit int) (string, error) {
-	submatches := _URLRegexp.FindAllStringSubmatch(repoUrl, -1)
-	if len(submatches) == 0 {
-		return "", fmt.Errorf("failed to match github repository: %w", unexpectedRepoUrlError)
-	}
-
-	matchesMap := map[string]string{}
-	for i, n := range submatches[0] {
-		matchesMap[_URLRegexpNames[i]] = n
+	owner, repo, err := f.parseOwnerAndRepoFromUrl(ctx, repoUrl)
+	if err != nil {
+		return "", fmt.Errorf("could not complile regex to check repoURl: %w", err)
 	}
 	lg := log.FromContext(ctx)
 	ghClient, err := f.ghClientBuilder.createAuthenticatedGhClient(ctx, token)
 	if err != nil {
 		return "", fmt.Errorf("failed to create authenticated GitHub client: %w", err)
 	}
-	file, dir, resp, err := ghClient.Repositories.GetContents(ctx, matchesMap["owner"], matchesMap["repo"], filepath, &github.RepositoryContentGetOptions{Ref: ref})
+	file, dir, resp, err := ghClient.Repositories.GetContents(ctx, owner, repo, filepath, &github.RepositoryContentGetOptions{Ref: ref})
 	if err != nil {
 		checkRateLimitError(err)
 		bytes, _ := io.ReadAll(resp.Body)
@@ -79,4 +74,24 @@ func (f downloadFileCapability) DownloadFile(ctx context.Context, repoUrl, filep
 		return "", fmt.Errorf("content reading error: %w", err)
 	}
 	return content, nil
+}
+
+func (f downloadFileCapability) parseOwnerAndRepoFromUrl(ctx context.Context, url string) (owner string, repo string, err error) {
+	urlRegexp, err := regexp.Compile(`(?Um)^` + regexp.QuoteMeta(f.githubBaseUrl) + `/(?P<owner>[^/]+)/(?P<repo>[^/]+)(/|(.git)?)$`)
+	if err != nil {
+		return "", "", fmt.Errorf("could not complile regex to check repoURl: %w", err)
+	}
+
+	urlRegexpNames := urlRegexp.SubexpNames()
+	matches := urlRegexp.FindAllStringSubmatch(url, -1)
+	if len(matches) == 0 {
+		return "", "", fmt.Errorf("failed to match github repository: %w", unexpectedRepoUrlError)
+	}
+
+	matchesMap := map[string]string{}
+	for i, n := range matches[0] {
+		matchesMap[urlRegexpNames[i]] = n
+	}
+	log.FromContext(ctx).V(logs.DebugLevel).Info("parsed values from GitHub URL", "owner", matchesMap["owner"], "repo", matchesMap["repo"])
+	return matchesMap["owner"], matchesMap["repo"], nil
 }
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

File: pkg/serviceprovider/github/downloadfilecapability.go [hunk 1]
```diff
@@ -80,3 +75,23 @@ func (f downloadFileCapability) DownloadFile(ctx context.Context, repoUrl, filep
 	}
 	return content, nil
 }
+
+func (f downloadFileCapability) parseOwnerAndRepoFromUrl(ctx context.Context, url string) (owner string, repo string, err error) {
+	urlRegexp, err := regexp.Compile(`(?Um)^` + regexp.QuoteMeta(f.githubBaseUrl) + `/(?P<owner>[^/]+)/(?P<repo>[^/]+)(/|(.git)?)$`)
```
metlos, line 80: the reqex could be precompiled  and stored as a field of the capability instead of the base url..

metlos (commented):
lgtm apart from 1 minor comment.

pkg/serviceprovider/github/downloadfilecapability.go
```diff
@@ -80,3 +75,23 @@ func (f downloadFileCapability) DownloadFile(ctx context.Context, repoUrl, filep
...
+	matches := urlRegexp.FindAllStringSubmatch(url, -1)
+	if len(matches) == 0 {
+		return "", "", fmt.Errorf("failed to match github repository: %w", unexpectedRepoUrlError)
+	}
+
+	matchesMap := map[string]string{}
+	for i, n := range matches[0] {
+		matchesMap[urlRegexpNames[i]] = n
+	}
+	log.FromContext(ctx).V(logs.DebugLevel).Info("parsed values from GitHub URL", "owner", matchesMap["owner"], "repo", matchesMap["repo"])
```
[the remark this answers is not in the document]
xbaran4: Good idea.

pkg/serviceprovider/github/downloadfilecapability.go
```diff
@@ -80,3 +75,23 @@ func (f downloadFileCapability) DownloadFile(ctx context.Context, repoUrl, filep
...
+func (f downloadFileCapability) parseOwnerAndRepoFromUrl(ctx context.Context, url string) (owner string, repo string, err error) {
+	urlRegexp, err := regexp.Compile(`(?Um)^` + regexp.QuoteMeta(f.githubBaseUrl) + `/(?P<owner>[^/]+)/(?P<repo>[^/]+)(/|(.git)?)$`)
+	if err != nil {
+		return "", "", fmt.Errorf("could not complile regex to check repoURl: %w", err)
+	}
+
+	urlRegexpNames := urlRegexp.SubexpNames()
+	matches := urlRegexp.FindAllStringSubmatch(url, -1)
+	if len(matches) == 0 {
+		return "", "", fmt.Errorf("failed to match github repository: %w", unexpectedRepoUrlError)
```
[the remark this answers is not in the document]
xbaran4: Good idea.

pkg/serviceprovider/github/downloadfilecapability.go
xbaran4: It is possible. I will define a constructor function for the capability struct.

pkg/serviceprovider/github/githubclientbuilder.go
```diff
@@ -50,5 +50,19 @@ func (g *githubClientBuilder) createAuthenticatedGhClient(ctx context.Context, s
 	ctx = context.WithValue(ctx, oauth2.HTTPClient, g.httpClient)
 	ts := oauth2.StaticTokenSource(&oauth2.Token{AccessToken: tokenData.AccessToken})
 	lg.V(logs.DebugLevel).Info("Created new github client", "SPIAccessToken", spiToken)
+
+	// We need the githubBaseUrl in the githubClientBuilder struct to decide whether to create
```
sparkoo, line 54: isn't it better to describe in an actual issue?

sparkoo (approved):

pkg/serviceprovider/github/githubclientbuilder.go
xbaran4: Yea, and we already made a decision about it, so I will adjust the comment.

pkg/serviceprovider/github/downloadfilecapability.go
```diff
@@ -42,25 +46,31 @@ var (
...
-var _URLRegexpNames = _URLRegexp.SubexpNames()
-
-func (f downloadFileCapability) DownloadFile(ctx context.Context, repoUrl, filepath, ref string, token *api.SPIAccessToken, maxFileSizeLimit int) (string, error) {
-	submatches := _URLRegexp.FindAllStringSubmatch(repoUrl, -1)
-	if len(submatches) == 0 {
-		return "", fmt.Errorf("failed to match github repository: %w", unexpectedRepoUrlError)
+func NewDownloadFileCapability(httpClient *http.Client, ghClientBuilder githubClientBuilder, baseUrl string) (serviceprovider.DownloadFileCapability, error) {
+	reg, err := regexp.Compile(`(?Um)^` + regexp.QuoteMeta(baseUrl) + `/(?P<owner>[^/]+)/(?P<repo>[^/]+)(/|(.git)?)$`)
+	if err != nil {
+		return nil, fmt.Errorf("compliling repoUrl matching regex for GitHub baseUrl %s failed with error: %w", baseUrl, err)
```
mshaposhnik, line 52:
```suggestion
		return nil, fmt.Errorf("compiling repoUrl matching regex for GitHub baseUrl %s failed with error: %w", baseUrl, err)
```

pkg/serviceprovider/github/downloadfilecapability.go
```diff
@@ -42,25 +46,31 @@ var (
...
+		ghClientBuilder: ghClientBuilder,
+		ghBaseUrl:       baseUrl,
+		ghUrlRegexp:     reg,
+	}, nil
+}
+
+func (d downloadFileCapability) DownloadFile(ctx context.Context, repoUrl, filepath, ref string, token *api.SPIAccessToken, maxFileSizeLimit int) (string, error) {
+	owner, repo, err := d.parseOwnerAndRepoFromUrl(ctx, repoUrl)
+	if err != nil {
+		return "", fmt.Errorf("could not parse repostitory name and owner from repoUrl: %w", err)
```
mshaposhnik, line 66:
```suggestion
		return "", fmt.Errorf("could not parse repository name and owner from repoUrl: %w", err)
```

mshaposhnik (approved):

metlos (approved):

pkg/serviceprovider/github/downloadfilecapability.go
xbaran4: fixed

pkg/serviceprovider/github/downloadfilecapability.go
xbaran4: fixed

PR #613, opened 2023-05-26
Redirect proxy support 

skabashnyuk:
### What does this PR do?
 - Add a notion of Redirect proxy
 - If the OAuth redirection proxy is configured - then the callback URL is changed to point the redirection proxy
 - To make the redirection proxy stateless the original callback is encoded with a state to the new state parameter.

n/a

https://issues.redhat.com/browse/SVPI-476
- allow using a single OAuth application to test SPI that can be deployed on a cluster with can have each time a new url.
- allow using a single redirection proxy for the multicluster setup.

### How to test this PR?
- deploy new images
``` 
image operator quay.io/skabashn/service-provider-integration-operator:redirectProxy_2023_05_26__10_20_42 
image OAuth quay.io/skabashn/service-provider-integration-oauth:redirectProxy_2023_05_26__10_20_42
```
- change configuration 
```
kubectl -n spi-system patch configmap spi-oauth-service-environment-config --type=merge -p '{"data": {"REDIRECT_PROXY_URL": "'https://localhost:8080/oauth/callback'"}}'
kubectl -n spi-system rollout restart deployment/spi-controller-manager deployment/spi-oauth-service


```


albarbaro:
Nice! All great and working for me.

mshaposhnik (approved):

cmd/oauth/oauthcli/cli.go
```diff
@@ -21,10 +21,11 @@ import (
...
+	ServiceAddr      string `arg:"--service-addr, env" default:"0.0.0.0:8000" help:"Service address to listen on"`
+	AllowedOrigins   string `arg:"--allowed-origins, env" default:"https://console.redhat.com,https://console.stage.redhat.com,https://console.dev.redhat.com,https://prod.foo.redhat.com:1337" help:"Comma-separated list of domains allowed for cross-domain requests"`
+	KubeConfig       string `arg:"--kubeconfig, env" default:"" help:""`
+	KubeInsecureTLS  bool   `arg:"--kube-insecure-tls, env" default:"false" help:"Whether is allowed or not insecure kubernetes tls connection."`
+	ApiServer        string `arg:"--api-server, env:API_SERVER" default:"" help:"host:port of the Kubernetes API server to use when handling HTTP requests"`
+	ApiServerCAPath  string `arg:"--ca-path, env:API_SERVER_CA_PATH" default:"" help:"the path to the CA certificate to use when connecting to the Kubernetes API server"`
+	RedirectProxyUrl string `arg:"--redirect-proxy-url, env:REDIRECT_PROXY_URL" default:"" help:"the URL of OAuth redirection proxy used in the tests to maintain predictable callback URL"`
```
metlos, line 30: Is this really only ever to be useful in the tests? Also, I would maybe be more explicit with the name and call it `OAuthRedirectProxyUrl`.

metlos (commented):
While I understand the intention, I'm still not completely sure this is the right way to approach the problem.

Essentially, I see 2 solutions to this:
1. The one offered here, where the service provider can have the proxy URL set as its redirect and the proxy will then do the final "hop" to the actual deployment URL of SPI.
2. Just use the BASE_URL that both the SPI operator and OAuth service already understand. This URL needs to be served by the OAuth service.

### Pros of proxy
The advantage of the first one is that we explicitly support dynamically deployed OAuth service which can be served by a proxy that can live statically and doesn't require any state.

### Cons of proxy
Because our tokens are not signed nor encrypted, an attacker can modify the state to point to a different "final hop" URL, which could for example be a phishing site where they could try to extract some data from the end user. Note that this attacker might on even have access to our systems. All they needs is an address that uses our proxy in the url and some clever way of making the user to click it thinking it is a legit link to our systems.

### Notes on proxy
The proxy makes it possible for the `/login` endpoint to be located on a different domain than the redirect URL configured on the service provider. This may or may not raise eyebrows.

### Pros of BASE_URL
All Oauth service endpoints live on the same domain. No additional state needed to transfer across the OAuth flow to be able to arrive at the correct final destination (which eliminates the above described attack vector).

### Cons of BASE_URL
The BASE_URL and the redirect URL in the service provider need to always match. But this can IMHO be achieved at the ingress level.

metlos:
After discussing the questions I raised in the review, we arrived at the following conclusions:

The storing of the redirect url in the state is required primarily because of the fact that we need a single Github OAuth application to serve multiple installations of RHTAP. This is currently mainly required by the end to end tests (of which there can be more running in parallel) but can also be required in the future (e.g. a single github app for both staging and prod installations of rhtap).

Therefore this somehow needs to work. We devised the following plan:

1. Merge this PR to unblock QE
2. Modify the proxy to only forward to configured domains (using regex matching for example)
3. If further tightening of security is required, we can think about signing the state we a secret shared between our OAuth service and the proxy. This is a little bit involved because of the need for secret rotation etc, so we're not going ahead with it at the moment, relying on the protection provided by the redirect url check instead.

metlos (approved):

PR #618, opened 2023-05-29
Simple version of target setting controller

mshaposhnik:
### What does this PR do?
Adds a controller which watches the SEB instances and adds targets to the RemoteSecrets which is matching by app/environment labels.

Known issues:
 - Handle the SEB deletion and delete the respective target (wip now)
 - Handle per-component secrets

https://issues.redhat.com/browse/SVPI-492

### How to test this PR?

 
 
 -  ~On bare SPI installations, like minikube, add the CRD-s for `SnapshotEnvironmentBinding` and `Environment`.
        Just c/p from there:   https://github.com/redhat-appstudio/application-api/tree/main/config/crd/bases
        Not need on stg.~
        
 -  Start runtime. Add the remote secret:

 ```
kind: RemoteSecret
metadata:
  name: test-remote-secret
  namespace: default
  labels:
    appstudio.application: so-sample
spec:
  secret:
    name: secret-from-remote
    type: Opaque
 ```
 
 - Add test environment:
 
```
kind: Environment
metadata:
  name: development
  namespace: default
  labels:
    toolchain.dev.openshift.com/owner: mshaposh
    toolchain.dev.openshift.com/provider: codeready-toolchain
spec:
  displayName: Development
  type: Non-POC

``` 

- Add SEB:

```
metadata:
  generateName: so-sample-development-binding-
  labels:
    appstudio.application: so-sample
  namespace: default
spec:
  application: so-sample
  components:
  - configuration:
      replicas: 1
    name: devfile-sample-go-basic-fwzb
  environment: development
  snapshot: so-sample-zqpld
```
 
 - Verify RemoteSecret have target added:
 
 ```
kind: RemoteSecret
metadata:
  name: test-remote-secret
  namespace: default
  labels:
    appstudio.application: so-sample
    ...
spec:
  secret:
    name: secret-from-remote
    type: Opaque
  targets:
    - namespace: default
status:
  conditions:
    - lastTransitionTime: '2023-05-29T09:29:09Z'
      message: ''
      reason: Valid
      ....
 ```
 

 - Remove SEB and verify RemoteSecret target has been deleted:
 
 ```
kubectl delete snapshotenvironmentbinding so-sample-development-binding-xxxxx -n default
 ```

```
```


controllers/snapshotenvironmentbinding_controller.go
```diff
@@ -0,0 +1,194 @@
...
+			return ctrl.Result{}, fmt.Errorf("failed to update secret: %w", err)
+		}
+	}
+
+	return ctrl.Result{}, nil
+}
+
+func addTargetIfNotExists(secret *api.RemoteSecret, target api.RemoteSecretTarget) {
+	for idx := range secret.Spec.Targets {
+		existingTarget := secret.Spec.Targets[idx]
```
gazarenkov, line 131: I think this kind of check is excessive, it is RemoteSecret's business to care about its integrity.

controllers/snapshotenvironmentbinding_controller.go
```diff
@@ -0,0 +1,194 @@
...
+			return ctrl.Result{}, nil
+		}
+
+		return ctrl.Result{}, fmt.Errorf("failed to get the SnapshotEnvironmentBinding: %w", err)
+	}
+
+	finalizationResult, err := r.finalizers.Finalize(ctx, &snapshotEnvBinding)
+	if err != nil {
+		// if the finalization fails, the finalizer stays in place, and so we don't want any repeated attempts until
+		// we get another reconciliation due to cluster state change
```
gazarenkov, line 79: If your goal is to not to requeue if finalization fails, I doubt it will work like this because "The Controller will requeue the Request to be processed again if an error is non-nil or	Result.Requeue is true" and in your case the error is non-nil

controllers/snapshotenvironmentbinding_controller.go
```diff
@@ -0,0 +1,194 @@
...
+		return ctrl.Result{Requeue: false}, fmt.Errorf("failed to finalize: %w", err)
+	}
+	if finalizationResult.Updated {
+		if err = r.k8sClient.Update(ctx, &snapshotEnvBinding); err != nil {
```
gazarenkov, line 83: Are we going to update the SED ?

controllers/snapshotenvironmentbinding_controller.go
metlos: Huh, you're right :) This has been copied from the remote secret controller that claims the same thing. I'm not sure about not wanting to retry though... Maybe we actually should retry, because the error condition may clean up. The retries are happening with exponential back-off so it's not like we're going to clog up the cluster.

controllers/snapshotenvironmentbinding_controller.go
metlos: The controllers should not update the spec, because the spec should express the user's intent. So the RS controller would just leave the duplicate entries in the spec and just document in the status that they are duplicates and that it is ignoring them.

In another words, I would keep this sanity check in place.

controllers/snapshotenvironmentbinding_controller.go
```diff
@@ -0,0 +1,194 @@
...
+		existingTarget := secret.Spec.Targets[idx]
+		// Add the rest of the fields there
+		if existingTarget.Namespace == target.Namespace {
+			secret.Spec.Targets = append(secret.Spec.Targets[:idx], secret.Spec.Targets[idx+1:]...)
+		}
+	}
+}
+
+func detectTargetFromEnvironment(ctx context.Context, environment appstudiov1alpha1.Environment) (api.RemoteSecretTarget, error) {
+	// Dummy load. Real impl should reverse pass Env -> DTC -> DT -> SpaceRequest to find out full target details
```
metlos, line 151: We should track this as an issue.

metlos (approved):

controllers/snapshotenvironmentbinding_controller.go
gazarenkov: @metlos I am not sure if it makes sense to collect unlimited amount of duplicated objects, it will look really confusing. User's intent is not changed, user wants to have a secret in the particular namespace and controller is trying to do that. As discussed, this array is a set by its nature because it is not possible to have the same object with the same name on the same namespace and I believe we rather need to find a way to keep its contract internally than ask the external code to care of it.

controllers/snapshotenvironmentbinding_controller.go
```diff
@@ -0,0 +1,194 @@
...
+		// we get another reconciliation due to cluster state change
+		return ctrl.Result{Requeue: false}, fmt.Errorf("failed to finalize: %w", err)
+	}
+	if finalizationResult.Updated {
+		if err = r.k8sClient.Update(ctx, &snapshotEnvBinding); err != nil {
+			return ctrl.Result{}, fmt.Errorf("failed to update snapshot env binding based on finalization result: %w", err)
+		}
+	}
+	if finalizationResult.StatusUpdated {
+		if err = r.k8sClient.Status().Update(ctx, &snapshotEnvBinding); err != nil {
```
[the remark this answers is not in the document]
mshaposhnik: no, we should not...

controllers/snapshotenvironmentbinding_controller.go
mshaposhnik: yes it must set finalizer name into metadata

controllers/snapshotenvironmentbinding_controller.go
```diff
@@ -0,0 +1,218 @@
...
+
+const (
+	ApplicationLabelName                   = "appstudio.application"
+	EnvironmentLabelName                   = "appstudio.environment"
+	ComponentLabelName                     = "appstudio.component"
+	linkedRemoteSecretsTargetFinalizerName = "spi.appstudio.redhat.com/remote-secrets" //#nosec G101 -- false positive, just label name
+)
+
+var (
+	unableToFetchRemoteSecretsError = stderrors.New("unable to fetch the Remote Secrets list")
```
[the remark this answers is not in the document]
mshaposhnik: fixed

PR #621, opened 2023-06-02
Remote secret repo usage

skabashnyuk:
### What does this PR do?
 - used https://github.com/redhat-appstudio/remote-secret as a remote-secret implementation
 

n/a

https://issues.redhat.com/browse/SVPI-494

### How to test this PR?
- `make deploy_minikube`
- ` quay.io/skabashn/service-provider-integration-operator:remote-secret-repo_2023_06_07__11_07_47`
- `quay.io/skabashn/service-provider-integration-oauth:remote-secret-repo_2023_06_07__11_07_47`
-  basic spi-test
- remote-secret happy path test


metlos (approved):
While I fully understand that this is the simplest way of splitting the remote secrets out of the SPI codebase, I'm not fully comfortable with the fact that all the utility classes and the `bindings` package from controllers is being moved to remote secrets, too. 

I needs to be moved for sure, but for example the `bindings` package feels weird in the new repository where it defines all the abstract interfaces to then only have a single implementation in the remote-secret repo. That just looks weird and would be IMHO better served by its own module. At the same time, I don't think that single package being shared between remote secrets and SPI warrants it to be moved to yet another repository, but it certainly feels like a separate module.

So for now, I don't want to block the move (which I think is in itself a good thing) but IMHO we need to work on proper code separation because we're slowly starting to make a mess in the codebase.

mshaposhnik (approved):

controllers/reconcile_error_test.go
xbaran4, line 1: Since `controllers/reconcile_error_test.go` is removed shouldn't this test file go as well?

docs/ADMIN.md
```diff
@@ -56,7 +56,7 @@ and oauth service to have the same value for them):
...
 | --config-file                                         | CONFIGFILE                     | /etc/spi/config.yaml     | The location of the configuration file.                                                                                                                                                                                            |
-| --spi-instance-id                                     | SPIINSTANCEID                  | spi-1                    | ID of this SPI instance. Used to avoid conflicts when multiple SPI instances uses shared resources (e.g. secretstorage).                                                                                                           |
+| --instance-id                                         | INSTANCEID                     | spi-1                    | ID of this SPI instance. Used to avoid conflicts when multiple SPI instances uses shared resources (e.g. secretstorage).                                                                                                           |
```
xbaran4, line 59: Should the description of the argument change as well? Can we call our whole system (service, spi operator, RS operator) spi ? If no, do we have some different naming?

xbaran4 (approved):
Just some nitpicks, otherwise we can merge. I agree with Lukas that we should think about defining some boundaries for what parts of the codebase are shared and what are separate otherwise it will be really hard to orientate in the code.

docs/ADMIN.md
xbaran4: Correct me if I am wrong. We are changing the name of this argument from `--spi-instance-id` to `--instance-id ` because now the RS-operator needs its own ID as well as SPI-operator.

The description says `ID of this SPI instance.` What is meant by `SPI instance` is it just SPI-operator or both SPI-operator and RS-operator?

xbaran4: Ok, I think I understand now, the problem was in my understanding. We can keep the description how as it is now.

PR #626, opened 2023-06-13
Add tags cleanup

mshaposhnik:
### What does this PR do?
${TITLE}

https://issues.redhat.com/browse/SVPI-501

### How to test this PR?

```
```


xbaran4 (commented):
This is a valid fix. But I would maybe prefer just removing the query parameters from the error response here: https://github.com/redhat-appstudio/service-provider-integration-operator/blob/main/oauth/handlers.go#L84

This means that if the query parameters are "ok" we put them into the template, and the HTML page returned is injection-safe (by html/template package). If the templating fails (which might mean that the query parameters are corrupt somehow), we just do not include the parameters in the response.

xbaran4 (approved):

PR #628, opened 2023-06-14
Make SEB finalizer do not fail when enviromnent is deleted

mshaposhnik:
### What does this PR do?

As there is no `owner ref` or `finalizer` between `Environment` and `SnapshotEnvironmentBinding`, the possible situation is when `Environment` is already deleted from the namespace, and `SnapshotEnvironmentBinding`  finalizer is failing when trying to delete.

This fix is simplifies the RemoteSecret target calculation in finalizer, to just pick-up same namespace as in SEB (which technically makes no difference as for now).  As for the future Namespace-backed environmets, this situation must be discussed 
to add some relations between `Environment` and `SnapshotEnvironmentBinding` to avoid such a erroneous state.

https://issues.redhat.com/browse/SVPI-513

### How to test this PR?

 
 Run cluster, add the first Environment and then the SEB, then remove them in same order. Must be no problems removing the SEB.
 
 Samples:
 
 ```
kind: Environment
metadata:
  name: development
  namespace: default
  labels:
    toolchain.dev.openshift.com/owner: mshaposh
    toolchain.dev.openshift.com/provider: codeready-toolchain
spec:
  displayName: Development
  type: Non-POC
 ```
 
 
 ```
metadata:
  labels:
    appstudio.application: so-sample
  name: so-sample-development-binding-l5btw
  namespace: default
spec:
  application: so-sample
  components:
  - configuration:
      replicas: 1
    name: devfile-sample-go-basic-fwzb
  environment: development
  snapshot: so-sample-zqpld
 ```

```
```


xbaran4 (commented):
LGTM, but could you show some samples of Environment and SEB for testing? Because I do not have much experience with them.

controllers/snapshotenvironmentbinding_controller.go
```diff
@@ -186,23 +185,17 @@ func (f *linkedRemoteSecretsFinalizer) Finalize(ctx context.Context, obj client.
...
 	}
 
-	// Get the Environment CR
-	environment := appstudiov1alpha1.Environment{}
-	err := f.client.Get(ctx, types.NamespacedName{Name: snapshotEnvBinding.Spec.Environment, Namespace: snapshotEnvBinding.Namespace}, &environment)
-	if err != nil {
-		return res, unableToFetchEnvironmentError
-	}
-
-	target, err := detectTargetFromEnvironment(ctx, environment)
```
metlos, line 196: Is this function still used somewhere else? Can we run into similar problems like in this finalizer in other workflows, too?

mshaposhnik: Yes it is used on target creation too, but hopefully the situation when SEB created w/out Env will never happen yet (but yes we still need to link them somehow to make 101% sure...)

metlos (approved):

xbaran4 (approved):

PR #635, opened 2023-06-26
Register remote CRDs into integration tests suite and try to use them

mshaposhnik:
### What does this PR do?
Registers remote CRDs for integration tests:
 - RemoteSecret
 - Environment
 - SnapshotEnvironmentBidning
 
 Added happy path test for target setting controller.

https://issues.redhat.com/browse/SVPI-523

### How to test this PR?

```
```

### DIFF
--- a/integration_tests/integrationtest.go
+++ b/integration_tests/integrationtest.go
@@ -17,11 +17,12 @@ package integrationtests
 import (
 	"context"
 	"fmt"
-	"github.com/redhat-appstudio/application-api/api/v1alpha1"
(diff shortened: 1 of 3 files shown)


metlos (approved):
This is the minimal approach to get this working with the current integration test infrastructure. However, I think that a more future-proof way of "fixing" this would be to try and use the https://github.com/metlos/crenv library that I adapted from this code and made it more generic such that we would not have to go through these exercises each time we need to add a new type of object to follow.

PR #701, opened 2023-09-12
Fix targets detection and iteration

mshaposhnik:
### What does this PR do?
Fix panic in targets cleanup fuction and add more accurate target calculation in cleanup.

https://issues.redhat.com/browse/SVPI-614

### How to test this PR?

```
```

### DIFF
--- a/controllers/snapshotenvironmentbinding_controller.go
+++ b/controllers/snapshotenvironmentbinding_controller.go
@@ -157,7 +157,7 @@ func addTargetIfNotExists(secret *rapi.RemoteSecret, target rapi.RemoteSecretTar
 }
 
 func removeTarget(secret *rapi.RemoteSecret, target rapi.RemoteSecretTarget) {
-	tmp := secret.Spec.Targets[:0]
+	tmp := make([]rapi.RemoteSecretTarget, 0, len(secret.Spec.Targets))
 	for _, existingTarget := range secret.Spec.Targets {
 		if !targetsMatch(existingTarget, target) {
 			tmp = append(tmp, existingTarget)
--- a/controllers/snapshotenvironmentbinding_controller.go
+++ b/controllers/snapshotenvironmentbinding_controller.go
@@ -213,6 +213,9 @@ func (f *linkedRemoteSecretsFinalizer) Finalize(ctx context.Context, obj client.
 		return finalizer.Result{}, fmt.Errorf("failed to load environment from cluster: %w", err)
 	}
 	target, err := detectTargetFromEnvironment(ctx, environment)
+	if err != nil {
+		return finalizer.Result{}, fmt.Errorf("failed to detect target from environment: %w", err)
+	}
 
 	for rs := range remoteSecretsList.Items {
 		remoteSecret := remoteSecretsList.Items[rs]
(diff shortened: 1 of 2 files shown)


controllers/snapshotenvironmentbinding_controller.go
```diff
@@ -157,12 +157,13 @@ func addTargetIfNotExists(secret *rapi.RemoteSecret, target rapi.RemoteSecretTar
 }
 
 func removeTarget(secret *rapi.RemoteSecret, target rapi.RemoteSecretTarget) {
-	for idx := range secret.Spec.Targets {
-		existingTarget := secret.Spec.Targets[idx]
-		if targetsMatch(existingTarget, target) {
-			secret.Spec.Targets = append(secret.Spec.Targets[:idx], secret.Spec.Targets[idx+1:]...)
+	tmp := secret.Spec.Targets[:0]
```
metlos, line 160: This is creating an empty slice backed by the `secret.Spec.Target` array. You then append to this slice which IMHO just creates a new array under the hood. So how about just creating a new array here with the same capacity as the targets?

E.g.:
```go
tmp := make([]rapi.RemoteSecretTarget, 0, len(secret.Spec.Targets))
```

This will avoid re-allocations of the `tmp` array (and will waste at most 1 element in it).

controllers/snapshotenvironmentbinding_controller.go
```diff
@@ -157,12 +157,13 @@ func addTargetIfNotExists(secret *rapi.RemoteSecret, target rapi.RemoteSecretTar
...
+	for _, existingTarget := range secret.Spec.Targets {
+		if !targetsMatch(existingTarget, target) {
+			tmp = append(tmp, existingTarget)
 		}
 	}
+	secret.Spec.Targets = tmp
 }
 
 func targetsMatch(target1, target2 rapi.RemoteSecretTarget) bool {
```
metlos, line 169: Btw I would argue that ClusterCredentialsSecret SHOULD NOT be part of the equality, because all it means is that the operator will use different credentials to create the same secret.

controllers/snapshotenvironmentbinding_controller.go
```diff
@@ -205,7 +206,16 @@ func (f *linkedRemoteSecretsFinalizer) Finalize(ctx context.Context, obj client.
 		return finalizer.Result{}, nil
 	}
 
-	target := rapi.RemoteSecretTarget{Namespace: snapshotEnvBinding.Namespace}
+	// Get the Environment CR
```
metlos, line 209: Is this change tested? I assume it is because of the possibility of the remote-cluster targets?

mshaposhnik: yep

controllers/snapshotenvironmentbinding_controller.go
mshaposhnik: ok

controllers/snapshotenvironmentbinding_controller.go
mshaposhnik: so the proposal is to exlude `ClusterCredentialsSecret` from targets comparing ?

xbaran4 (approved):

PR #705, opened 2023-09-13
feat: SPIFileContentRequest support for RemoteSecret

xbaran4:
### What does this PR do?
Enables SPIFileContentRequest to use credentials from RemoteSecret. It uses the same search algorithm for RemoteSecrets used in SPIAccessCheck (read more [here](https://github.com/redhat-appstudio/service-provider-integration-operator/pull/691) https://github.com/redhat-appstudio/service-provider-integration-operator/pull/691)

The reconcile logic is changed so that before creating binding and waiting for token, it searches for RemoteSecret it could use. If any is found, it uses the credentials and delivers the content. This fully preserves backward compatibility but complicates the reconciliation. Check #706 for version breaking the compatibility.

#### Negatives of current implementation
Because the logic is happening mostly on the reconcile level, where the fileDownload is initiated, the reconciler needs to access the token storage and also needs access to the RemoteSecret search, which is done two levels down in the GenericLookup.

#### More context
The idea was to make the SPIFileContentRequest behave similarly to SPIAccessCheck. However, they are different beasts. Most of the logic in SPIAccessCheck happens in the service provider's CheckRepositoryAccess method, whereas most of SPIFileContentRequest's logic happens in the reconcile. SPIFileContentRequest creates a SPIAccessTokenBinding and yields the token lookup to that binding. After an SPIAccessToken is linked to a binding, the SPIFileContentRequest uses this token to download the file content.

Consequently, one can only download a file with a token, even if it is a public repo. @mshaposhnik and @metlos pointed out that one of the reasons for this is that the GitHub API rate limit without a token is quite strict, and we would quickly hit it. In SPIAccessCheck, we avoid this rate limit for public repositories because we do a GET request on the repo website, not the API. (Maybe we could do something similar with SPIFileContentRequest???)

### How to test this PR?

```
```

### DIFF
--- a/api/v1beta1/spifilecontentrequest_types.go
+++ b/api/v1beta1/spifilecontentrequest_types.go
@@ -14,7 +14,9 @@
 
 package v1beta1
 
-import metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
 
 type SPIFileContentRequestSpec struct {
 	// FilePath defines target file path inside repository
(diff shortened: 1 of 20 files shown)


xbaran4:
Closed in favor of #706

PR #706, opened 2023-09-14
feat: SPIFileContentRequest support for RemoteSecret - version 2

xbaran4:
### What does this PR do?
Enables SPIFileContentRequest to use credentials from RemoteSecret. It uses the same search algorithm for RemoteSecrets used in SPIAccessCheck (read more https://github.com/redhat-appstudio/service-provider-integration-operator/pull/691)

The reconciliation logic is changed so that no binding is created. If there is a SPIAccessToken or RemoteSecret from which Credentials can be used, the SPIFileContentRequest uses them to download the file.

#### Pros
Massively simplified reconciliation (just check if service provider has downloadFileCapability and if so download the file).
Lookup in one consolidated place.

#### Cons
Breaks backward compatibilty: no bindings are created or used, no waiting for binding to be ready, removed fields from SPIFileContentRequest related to binding.

### How to test this PR?

1. Deploy SPI with SPIO img: `quay.io/pbaran/spis:SVPI-565-version2`
2. Deploy RemoteSecrets
3. Create PAT for GitHub with at least `repo` scope. (https://github.com/settings/tokens)

#### Using credentials from SPIAccessToken
1. Create SPIAccessTokenBinding (this creates SPIAccessToken):
```shell
kind: SPIAccessTokenBinding
metadata:
  name: binding-sa
spec:
  permissions:
    required:
      - type: r
        area: repository
  repoUrl: https://github.com/xbaran4/testing # your private repo
  secret:
    name: token-sa

EOF
```
2. Upload your PAT through the uploadUrl from the status.
3. Create SPIFIleContentRequest, replace the spec for your (may be private) repo file:
```shell
kind: SPIFileContentRequest
metadata:
  name: file-content-request
  namespace: default
spec:
  repoUrl: https://github.com/redhat-appstudio/service-provider-integration-operator # replace with your repo
  filePath: sample/filecontentrequest.yaml #replace with your filepath
EOF
```
4. Check that the phase of SPIFileContentRequest is delivered and that the file's content is base64 encoded in its status.

#### Using credentials from RemoteSecret
*If you followed the previous section, remove the SPIAccessToken*

1. Create RemoteSecret with your github PAT:
```shell
kind: RemoteSecret
metadata:
  name: myrs1
  namespace: default
  labels:
    appstudio.redhat.com/sp.host: github.com
spec:
  secret:
    generateName: secret-from-remote-
  targets:
    - namespace: default
data:
  password: # your GITHUB PAT
EOF
```
2. Continue with steps 3. and 4. from the previous section.

```
```


pkg/serviceprovider/github/downloadfilecapability.go
```diff
@@ -57,27 +60,37 @@ func NewDownloadFileCapability(httpClient *http.Client, ghClientBuilder githubCl
...
-	owner, repo, err := d.parseOwnerAndRepoFromUrl(ctx, repoUrl)
+func (d downloadFileCapability) DownloadFile(ctx context.Context, cl client.Client, request *api.SPIFileContentRequest, maxFileSizeLimit int) (string, error) {
+	owner, repo, err := d.parseOwnerAndRepoFromUrl(ctx, request.RepoUrl())
 	if err != nil {
 		return "", fmt.Errorf("could not parse repository name and owner from repoUrl: %w", err)
 	}
 	lg := log.FromContext(ctx)
-	ghClient, err := d.ghClientBuilder.createAuthenticatedGhClient(ctx, token)
+
+	credentials, err := d.lookup.LookupCredentials(ctx, cl, request)
```
metlos, line 74: What about inverting this and providing the credentials to this method? That way we can do away with the heavy `lookup` in the downloadCapability.

controllers/spifilecontentrequest_controller.go
```diff
@@ -172,110 +164,41 @@ func (r *SPIFileContentRequestReconciler) Reconcile(ctx context.Context, req ctr
...
+			r.updateFileRequestStatusError(ctx, &request, noSuitableServiceProviderFound)
 		}
 		return ctrl.Result{}, nil
 	}
-	//binding not yet fully ready to work with, let's try next time
-	if associatedBinding.Status.Phase == "" || associatedBinding.Status.UploadUrl == "" {
-		return reconcile.Result{}, nil
+	if sp.GetDownloadFileCapability() == nil {
+		r.updateFileRequestStatusError(ctx, &request, serviceprovider.FileDownloadNotSupportedError{})
+		return ctrl.Result{}, serviceprovider.FileDownloadNotSupportedError{}
```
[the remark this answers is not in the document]
xbaran4: This part is actually from the original reconcile logic, but you made a good point, and I will fix it.

pkg/serviceprovider/downloadfilecapability.go
```diff
@@ -27,7 +29,13 @@ func (f FileDownloadNotSupportedError) Error() string {
 	return "provided repository URL does not supports file downloading"
 }
 
+type NoCredentialsFoundError struct{}
+
+func (f NoCredentialsFoundError) Error() string {
+	return "no suitable credentials found for the SPIFileContentRequest"
```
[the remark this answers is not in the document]
xbaran4: Agreed.

controllers/spifilecontentrequest_controller.go
```diff
@@ -172,110 +164,41 @@ func (r *SPIFileContentRequestReconciler) Reconcile(ctx context.Context, req ctr
...
-			return ctrl.Result{}, fmt.Errorf("unable to fetch the SPI Access token: %w", err)
-		}
-		contents, err := sp.GetDownloadFileCapability().DownloadFile(ctx, request.Spec.RepoUrl, request.Spec.FilePath, request.Spec.Ref, token, r.Configuration.MaxFileDownloadSize)
-		if err != nil {
-			r.updateFileRequestStatusError(ctx, &request, err)
-			return reconcile.Result{}, fmt.Errorf("error fetching file content: %w", err)
+	contents, err := sp.GetDownloadFileCapability().DownloadFile(ctx, r.K8sClient, &request, r.Configuration.MaxFileDownloadSize)
+	if err != nil {
+		if stderrors.As(err, &serviceprovider.NoCredentialsFoundError{}) {
+			request.Status.Phase = api.SPIFileContentRequestPhaseAwaitingTokenData
```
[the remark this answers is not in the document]
xbaran4: I agree. We can probably remove the `AwaitingTokenData` phase with this new implementation. But we have no way of providing some similar feedback that there is no TokenData. In my current implementation, I am classifying it (a situation when no credentials are found) into the error phase.

pkg/serviceprovider/github/downloadfilecapability.go
xbaran4: Yes, I think that is a cleaner solution, although we will need to extend the serviceProvider interface to LookupCredentials.

xbaran4:
The E2E test (https://github.com/redhat-appstudio/e2e-tests/blob/main/tests/spi/get-file-content.go) is expected to fail as this change is backward incompatible.

xbaran4:
To adapt the test to the new logic should not be hard, but I assume we need to: disable the E2E test -> merge this PR -> adapt E2E test and merge new E2E test PR.

If you know of another way please let me know.

controllers/spifilecontentrequest_controller.go
```diff
@@ -135,23 +82,6 @@ func (r *SPIFileContentRequestReconciler) Reconcile(ctx context.Context, req ctr
 		return ctrl.Result{}, fmt.Errorf("failed to read the object: %w", err)
 	}
 
-	finalizationResult, err := r.finalizers.Finalize(ctx, &request)
```
mshaposhnik, line 138: maybe it makes sense to cleanup finalizers from existing objects?
like here:
https://github.com/redhat-appstudio/service-provider-integration-operator/pull/716/files#diff-2c65b925cdcacc42c273f6ac66fc1fa7d51333eca6ab564ee1a100553ea4133bR122-R126

mshaposhnik (approved):

xbaran4:
I was able to verify that HAS can use the new SPIFileContentRequest to download a devfile.
image
(the error is there because of the format of devfile but important for us is that the devfile's content was delivered)

controllers/spifilecontentrequest_controller.go
xbaran4: agreed. done

PR #716, opened 2023-09-22
Make RS or targets cleanup on demand by Environment or Application deletion event

mshaposhnik:
### What does this PR do?
Perform RS tagret cleanups by the one of following actions:
- Environment deleted (cleanup targets for secrets labelled with given env name)
- Application deleted (cleanup targets for secrets labelled with given app name)

https://issues.redhat.com/browse/SVPI-609
https://issues.redhat.com/browse/SVPI-616

### How to test this PR?

 Use minikube;

 **Experiment 1:**
 
 - Create RS, ENV, and SEB:
 ```
kind: RemoteSecret
metadata:
  name: demo-secret
  namespace: default
  labels:
    appstudio.redhat.com/application: rs-sample
    appstudio.redhat.com/environment: development
spec:
  secret:
    name: demo-secret-from-remote
    type: Opaque

 ```
  +
 ```
kind: Environment
metadata:
  name: development
  namespace: default
spec:
  displayName: rhpds
  tags:
    - managed
  configuration:
    env:
      - name: MY_FAKE_ENV_VAR
        value: "100"
  unstableConfigurationFields:
    clusterType: OpenShift
    kubernetesCredentials:
      allowInsecureSkipTLSVerify: true
      apiURL: 'https://api.cluster-vj5m7.vj5m7.sandbox687.opentlc.com:6443'
      clusterCredentialsSecret: env-cluster-credentials-rnnhh
      namespaces:
        - default
      targetNamespace: default
```
+
```
metadata:
  labels:
    appstudio.application: rs-sample
  name: seb-development
  namespace: default
spec:
  application: rs-sample
  components:
    - configuration:
        replicas: 0
      name: devfile-sample-go-basic
  environment: development
  snapshot: rs-sample-c6gb9
```
 
- Vefiry target added;
- Delete environment;
- Verify target deleted;

 **Experiment 2:**
  - Create RS, APP, ENV, and SEB:
 ```
kind: RemoteSecret
metadata:
  name: demo-secret
  namespace: default
  labels:
    appstudio.redhat.com/application: rs-sample
    appstudio.redhat.com/environment: development
spec:
  secret:
    name: demo-secret-from-remote
    type: Opaque

 ```
 +
 ```
 kind: Application
metadata:
  annotations:
    application.thumbnail: '9'
    finalizeCount: '0'
  resourceVersion: '679282354'
  name: rs-sample
  namespace: default
  finalizers:
spec:
  appModelRepository:
    url: 'https://github.com/redhat-appstudio-qe/java-sample'
  displayName: rs-sample
  gitOpsRepository:
    url: ''
 ```
  +
 ```
kind: Environment
metadata:
  name: development
  namespace: default
spec:
  displayName: rhpds
  tags:
    - managed
  configuration:
    env:
      - name: MY_FAKE_ENV_VAR
        value: "100"
  unstableConfigurationFields:
    clusterType: OpenShift
    kubernetesCredentials:
      allowInsecureSkipTLSVerify: true
      apiURL: 'https://api.cluster-vj5m7.vj5m7.sandbox687.opentlc.com:6443'
      clusterCredentialsSecret: env-cluster-credentials-rnnhh
      namespaces:
        - default
      targetNamespace: default
```
+
```
metadata:
  labels:
    appstudio.application: rs-sample
  name: seb-development
  namespace: default
spec:
  application: rs-sample
  components:
    - configuration:
        replicas: 0
      name: devfile-sample-go-basic
  environment: development
  snapshot: rs-sample-c6gb9
```
 
- Vefiry target added;
- Delete application;
- Verify remote secret deleted;
 

```
```


controllers/application_controller.go
```diff
@@ -0,0 +1,115 @@
...
+	}
+
+	if len(remoteSecretsList.Items) == 0 {
+		return finalizer.Result{}, nil
+	}
+
+	for rs := range remoteSecretsList.Items {
+		remoteSecret := remoteSecretsList.Items[rs]
+		applicationInSecret, set := remoteSecret.Labels[ApplicationLabelName]
+		if set && applicationInSecret != "" && applicationInSecret != application.Name {
```
metlos, line 104: This can just be `application.Name != applicationInSecret`, IMHO. 

The default value returned from the `remoteSecret.Labels[ApplicatonLabelName]` is empty string. So we don't need the `set` check and we also don't need the `applicationInSecret != ""` because we catch the same thing with `application.Name != applicationInSecret` because it can be safely assumed that `appllication.Name` is never empty in a finalizer (because k8s object's name cannot be empty once it is actually created).

controllers/application_controller.go
```diff
@@ -0,0 +1,115 @@
...
+			// this secret is intended for another application, bypassing it
+			continue
+		}
+		// cleanup the targets
+		remoteSecret.Spec.Targets = []rapi.RemoteSecretTarget{}
```
metlos, line 109: Wouldn't it be more logical to actually remote the remote secret? The application that it belongs to is being deleted, so what purpose would the RS serve after that? Do we want to keep it around just in case someone wanted to reuse it in another application? How would that work?

metlos (changes requested):
I don't like the fact that you hacked the SEB controller to reconcile both the SEB and the Environments. IMHO, the modifications to the environments should be done in a separate controller, similarly to how you handle the application.

Running the logic to update the environment with a finalizer and/or finalize the environment on each reconcile of a SEB just seems wrong to me. It would be much cleaner if you did that in a separate environment controller and IMHO it would also make the seb controller logic easier.

controllers/application_controller.go
mshaposhnik: I was thinking of this, the only concern was the accidental loss of secret data, but i thing we can go with it unless anyone complain :)

mshaposhnik:
@metlos the reason why it is done via SEB is the timing, we cannot add  target + finalizer when the ENV is created because the'y re can be added independently, without any relation to the application creation or startup, while we want to inject targets only when the app started deploying.

metlos:
> @metlos the reason why it is done via SEB is the timing, we cannot add target + finalizer when the ENV is created because the'y re can be added independently, without any relation to the application creation or startup, while we want to inject targets only when the app started deploying.

So you optimize for the case where no application is ever started in an environment by not putting the finalizer on the env in that case. Because as soon as there is at least 1 seb, we put the finalizer on the env and we never remove it again as far as I looked.

I'm not sure that is worth the complications in our codebase. I would much rather just have a mostly no-op environment reconciler and a simpler seb reconciler than 1 thing that tries to combine both.

Btw. I am not 100% sure the timing is always going to be ok (because I am mainly not sure how watches work in Kubernetes in enough detail). You basically transform a delete event of an environment into a series of reconcile requests to seb. Are you 100% sure that the delete of the env won't ever complete before all the reconciles of the corresponding sebs are done? because that's what the code relies on but I honestly can't say I know the answer to this with 100% certainty. 

If this was an ordinary reconciler directly watching and reconciling the envs, we can be quite sure all works as it should.

controllers/application_controller.go
metlos: Well, I think we should be cautious here and actually care more about accidental secret data retention, which is what happens here - these remote secret just hold on to data and are no longer reachable from anywhere.

controllers/application_controller.go
```diff
@@ -0,0 +1,122 @@
...
+	for rs := range remoteSecretsList.Items {
+		remoteSecret := remoteSecretsList.Items[rs]
+		applicationInSecret := remoteSecret.Labels[ApplicationLabelName]
+		if applicationInSecret != application.Name {
+			// this secret is intended for another application, bypassing it
+			continue
+		}
+		// remove the secret
+		if err := f.client.Delete(ctx, &remoteSecret); err != nil {
+			return finalizer.Result{}, unableToDeleteRemoteSecret
```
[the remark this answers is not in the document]
mshaposhnik: \+

controllers/environment_controller.go
```diff
@@ -0,0 +1,138 @@
...
+		return finalizer.Result{}, fmt.Errorf("failed to detect target from environment: %w", err)
+	}
+
+	for rs := range remoteSecretsList.Items {
+		remoteSecret := remoteSecretsList.Items[rs]
+		if !remoteSecretApplicableForEnvironment(remoteSecret, environment.Name) {
+			// this secret is intended for another environment, bypassing it
+			continue
+		}
+		removeTarget(&remoteSecret, target)
```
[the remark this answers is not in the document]
mshaposhnik: \+

metlos (approved):
LGTM with just a single minor nitpick.

PR #720, opened 2023-09-27
Enable RemoteSecret integration by default

skabashnyuk:
### What does this PR do?
 - {SUBJ}
 - updated documentation, added missing parameters 

n/a

This integration becomes a natural non-separated part. There is no sense in keeping it disabled.

### How to test this PR?
 - Generic tests


metlos (approved):
While the change itself is good, I wonder if we should deprecate this flag and later replace it with something different or drop it altogether. The only thing it seems to govern is the seb controller, while other things, like remote secret support in accesscheck, ignore it.

So, if we want to go down that route, I'd suggest renaming it to "enable gitops integration" or something like that to better express what it actually influences.

But frankly, I think we could easily do it altogether.

mshaposhnik (approved):

PR #730, opened 2023-10-04
Extra check if response is not empty before reading

skabashnyuk:
### What does this PR do?
{SUBJ}

n/a

https://issues.redhat.com/browse/SVPI-636

### How to test this PR?
1. Create SPIFileContentRequest
2. Create RemoteSecret with data that contains new lines in username and password.
3. Expected error in the operator logs


xbaran4 (approved):
LGTM

xbaran4 (commented):
I was thinking about the topic of reducing the amount of false positive errors and I think this place is where we could improve somehow (not in this PR). The tricky thing is to differentiate which error code indicates the fault of the controller and which of the user. Because 404 status code might mean that the user inputed a wrong URL, credentials, etc. or that the controller mismatched the credentials or a different controller bug.

I'm putting this comment here just to trigger thoughts/discussion.

PR #765, opened 2023-10-26
Add a flag to disable HTTP/2 in the webhooks and kube-rbac-proxy

metlos:
### What does this PR do?
`$TITLE`

https://issues.redhat.com/browse/SVPI-659

### How to test this PR?
See https://github.com/redhat-appstudio/remote-secret/pull/193

Test images: 
`quay.io/lkrejci/service-provider-integration-operator:svpi-659`
`quay.io/lkrejci/service-provider-integration-oauth:svpi-659`


cmd/cli.go
```diff
@@ -23,5 +25,11 @@ import (
 type CommonCliArgs struct {
 	rcmd.CommonCliArgs
 	rcmd.LoggingCliArgs
-	BaseUrl string `arg:"--base-url, env" help:"The externally accessible URL on which the OAuth service is listening. This is used to construct manual-upload and OAuth URLs"`
+	BaseUrl      string `arg:"--base-url, env" help:"The externally accessible URL on which the OAuth service is listening. This is used to construct manual-upload and OAuth URLs"`
+	DisableHTTP2 bool   `arg:"--disable-http2, env" default:"true" help:"whether to disable access using the HTTP/2 protocol."`
```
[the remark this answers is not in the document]
metlos: I was thinking about that but that would mean tying the remote secrets and spi together a little bit more. For example, I think the `ExposeProfiling` was a mistake because it is actually not supported in spi operator. So I'd rather duplicate a little bit of code to not tie the operators together more than save a couple of lines of code.

config/manager/manager.yaml
```diff
@@ -66,6 +66,7 @@ spec:
           - "--upstream=http://127.0.0.1:8080/"
           - "--logtostderr=true"
           - "--v=10"
+          - "--http2-disable=true"
```
mshaposhnik, line 69: Doesn't seem to match with CLI arg name

config/oauth/deployment.yaml
```diff
@@ -65,6 +65,7 @@ spec:
             - "--upstream=http://127.0.0.1:8080/"
             - "--logtostderr=true"
             - "--v=10"
+            - "--http2-disable=true"
```
mshaposhnik, line 68: Doesn't seem to match with CLI arg name

mshaposhnik (approved):

config/manager/manager.yaml
metlos: that's because this is how the argument is called in kube-rbac-proxy for which this argument is for.

config/oauth/deployment.yaml
metlos: same as above, this is for kube-rbac-proxy where the arg is called this name.

PR #777, opened 2023-11-03
Prepare tests for Pact

Katka92:
### What does this PR do?
Separating test env creation so it can be reused by Pact contract tests. No functionality should be changed so far.

https://issues.redhat.com/browse/HAC-5440

### How to test this PR?
Simply run `make test` to verify nothing has broken.


albarbaro:
looks all good to me

Katka92:
Hello @skabashnyuk , that's a good point. If it would be better to rename it to <something>_test.go, I can do it. I just need to call the method somehow from another package.
If you're interested in how it would be used, you can take a look at HAS repo, where it is already done the same way. [Here](https://github.com/redhat-appstudio/application-service/blob/main/contracts/application_pact_test.go#L88) how Pact tests are calling it, [here](https://github.com/redhat-appstudio/application-service/blob/main/controllers/start_test_env.go#L55) how they set it in unit tests and [here](https://github.com/redhat-appstudio/application-service/blob/main/controllers/suite_test.go#L38) the code itself.

